Issue #13, opened 2020-01-14
Memory Leak in JNI Save Index Function


jmazanec15:
In the JNI SaveIndex function, it calls clear on a vector of heap allocated objects. Clear calls the destructor of each object but does not delete the space of the actual object. This leads to a leak in memory.

https://github.com/opendistro-for-elasticsearch/k-NN/blob/10f825a900ed5e0cd499b36053d07321b0387a12/jni/src/v1736/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp#L113

PR #14, opened 2020-01-14
fixed memory leak in saveIndex

jmazanec15:
#13 

Added a for loop to delete each object in saveIndex function to prevent leakage in memory. Built libraries with the change.


vamshin (approved):
Thanks for the fix.

vamshin (commented):
Please raise the pull request against master?

PR #24, opened 2020-01-31
Various clean up around race conditions and exception handling

jschmitz28:
### Motivation
While running some load tests using the KNN Plugin, we encountered an Elasticsearch node crash. We don't have access to the crash logs on our end, but during the event we were running around 5,000 queries per second to various (on the order of a couple hundred) KNN indexes in the cluster, as well as running workflows which actively restore pre-built KNN indexes onto the cluster, run search queries against those indexes, and delete indexes which are no longer necessary (which also receive query traffic right up until the point at which they are deleted). 

### Description of changes

1. Remove all calls from c++ to KNN Index java, instead having init() return the index pointer and queryIndex() and gc() accept the index pointer
2. Add read write lock to KNN index java to allow parallel queries but not while index is being closed (which applies a write lock). Upon proceeding with the read lock, queries throw an IOException if the index has been closed while they were waiting.
3. Simplify exception semantics so that unexpected failure modes (e.g. bad alloc due to OOM) are left unhandled and lead to expected process crashes, while expected scenarios (execution of a query just after gc) leads to a proper checked exception (IOException)
4. Make sure call to gc() occurs first in case there's some exception thrown in onRemoval code prior to it

### DIFF
--- a/jni/src/v1736/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp
+++ b/jni/src/v1736/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp
@@ -184,7 +184,7 @@ JNIEXPORT jobjectArray JNICALL Java_com_amazon_opendistroforelasticsearch_knn_in
     return NULL;
 }
 
-JNIEXPORT jlong JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex_init(JNIEnv* env, jobject indexObject, jstring indexPath)
+JNIEXPORT jlong JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex_init(JNIEnv* env, jstring indexPath)
 {
     Space<float>* space = NULL;
     ObjectVector* dataset = NULL;
@@ -200,11 +200,11 @@ JNIEXPORT jlong JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v17
         env->ReleaseStringUTFChars(indexPath, indexString);
         has_exception_in_stack(env);
 
-        return (jlong) index;
-
         // free up memory
         delete space;
         delete dataset;
+
+        return (jlong) index;
     }
     catch (...) {
         if (space) { delete space; }
--- a/jni/src/v1736/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp
+++ b/jni/src/v1736/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp
@@ -91,8 +91,8 @@ JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v173
             dataset.push_back(new Object(object_ids[i], -1, env->GetArrayLength(vectorArray)*sizeof(float), vector));
             env->ReleaseFloatArrayElements(vectorArray, vector, 0);
         }
+        // free up memory
         env->ReleaseIntArrayElements(ids, object_ids, 0);
-
         index = MethodFactoryRegistry<float>::Instance().CreateMethod(false, "hnsw", "l2", *space, dataset);
 
         int paramsCount = env->GetArrayLength(algoParams);
@@ -106,7 +106,7 @@ JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v173
 
         index->CreateIndex(AnyParams(paramsList));
         has_exception_in_stack(env);
-        const char * indexString = env->GetStringUTFChars(indexPath, 0);
+        const char *indexString = env->GetStringUTFChars(indexPath, 0);
         index->SaveIndex(indexString);
         env->ReleaseStringUTFChars(indexPath, indexString);
         has_exception_in_stack(env);
@@ -114,7 +114,7 @@ JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v173
         // Free each object in the dataset. No need to clear the vector because it goes out of scope
         // immediately
         for (auto it = dataset.begin(); it != dataset.end(); it++) {
-            delete *it;
+             delete *it;
         }
         delete index;
         delete space;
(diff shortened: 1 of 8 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -104,8 +105,10 @@ public synchronized void rebuild() {
      *
      * @param removalNotification key, value that got evicted.
      */
-    public void onRemoval(RemovalNotification<String, KNNIndex> removalNotification) {
-        try {            KNNIndex knnIndex = removalNotification.getValue();
+    private void onRemoval(RemovalNotification<String, KNNIndex> removalNotification) {
+            KNNIndex knnIndex = removalNotification.getValue();
```
vamshin, line 109: We should have exception handling here for other errors. If the segmentation faults occur, it does crash the process as we do not handle sigsegv signal

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/v1736/KNNIndex.java
```diff
@@ -61,25 +60,38 @@ public long getIndexSize() {
...
-                setIndexSize(0);
-            } else {
-                setIndexSize(file.length()/1024 + 1); // convert to KB and round up
+    public KNNQueryResult[] queryIndex(final float[] query, final int k, final String[] algoParams) throws IOException {
+        Lock readLock = readWriteLock.readLock();
+        readLock.lock();
+
+        try {
+            if (this.isClosed) {
+                throw new IOException("Index is already closed");
```
vamshin, line 69: Should we move this to top of the function to avoid acquiring lock in first place ?

jni/src/v1736/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp
```diff
@@ -202,10 +199,8 @@ JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v173
...
         env->ReleaseStringUTFChars(indexPath, indexString);
         has_exception_in_stack(env);
-        jclass indexClass = env->GetObjectClass(indexObject);
-        jmethodID setIndex = env->GetMethodID(indexClass, "setIndex", "(J)V");
-        env->CallVoidMethod(indexObject, setIndex, (jlong)index);
-        has_exception_in_stack(env);
+
+        return (jlong) index;
 
         // free up memory
```
jmazanec15, line 203: This code will not be reached

jmazanec15 (commented):
Thanks for the contribution! Just 1 comment in the code. Additionally, would it be possible to either create an issue or add a description of the race condition and the failure that occurred because it?

jni/src/v1736/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp
jschmitz28: Oops, you're right - I haven't been able to get any of this code properly tested, intent of the PR was to send to Vamshi some changes to consider in debugging race conditions and he had offered to take a look. Next version of the PR fixes ordering here and also fixes a bad type signature in init()

jschmitz28: Fixed in https://github.com/opendistro-for-elasticsearch/k-NN/pull/24/commits/84eb45c0a0d560c85d622eecfe74dc9b644feb08

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/v1736/KNNIndex.java
jschmitz28: Per our call, the check as to whether the index has been closed needs to happen after acquisition of the read lock, which means that the write lock is not actively held. Without that, isClosed could be false, but then upon acquisition of the read lock it could've flipped to true.

jschmitz28: https://github.com/opendistro-for-elasticsearch/k-NN/pull/24/commits/112b83c38ab254f86b20c737ca403021ed001df6 fixes issue I mentioned on the call where write lock wasn't actually being acquired

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
jschmitz28: As long as executor.execute() happens no matter what, I don't have much opinion (or knowledge) about the other parts. I do think it's critical that the submission of the gc() happens regardless of whether the other code throws, so if you think it's appropriate to wrap the rest of the code in try/catch (any exception), then I'd probably just suggest making sure to submit the gc() call first.

vamshin:
Thanks for the PR. I will send another PR with following changes

1.  Merge the current PR's code changes on top of master 
2. Update  libKNNINdex.so with the jni changes
3. Make sure build and integ tests pass

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
vamshin: My point is ```gc()``` can fail when we try to delete the memory not in use which would raise signal from the C library.  In this case whether gc() is called inside or outside try/catch it would have no effect.  

Agree on submitting the gc() call first

vamshin (approved):

jni/src/v1736/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp
jschmitz28: > Additionally, would it be possible to either create an issue or add a description of the race condition and the failure that occurred because it?

I don't have details on the crash itself because I don't have access to the nodes which crashed. I added a "Motivation" section to the top of this PR to describe the circumstances that led to the crash.

Issue #32, opened 2020-02-03
JNI Initalize Library called every time graph init'd


jmazanec15:
Every time the init function is called, the initLibrary function is called as well. initLibrary should only be called once. It alters state so this could affect query threads.

PR #33, opened 2020-02-03
Added initLibrary function to JNI 

jmazanec15:
*Issue #, if available:*
#32 

*Description of changes:*
Added JNI function "initLibrary" to be called when the library is loaded. This call is only supposed to be made once. Built .so and .jnilib's. ./gradlew build passes and manual sanity checks pass.

https://github.com/nmslib/nmslib/blob/master/similarity_search/include/init.h#L18


vamshin (approved):
LGTM! Thanks for fixing

zhanghg08 (approved):
LGTM

PR #36, opened 2020-02-04
Fix a leak where FileWatchers are added but never removed

jschmitz28:
### Description

Fix a leak where FileWatchers are added but never removed upon being expired from the cache, and some minor updates to privatize the cache by calling the public API in consumers

Prior to this change, FileWatchers to track when a KNN index is deleted are added to the ResourceWatcherService but never removed even after indexes have been deleted, which causes the set of watchers to grow unbounded over time. This change updates the logic to clean up FileWatchers upon cache eviction.


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -175,18 +178,57 @@ public void setCacheCapacityReached(Boolean value) {
...
+        // TODO verify that this is safe - ideally we'd explicitly ensure that the FileWatcher is only checked
+        // after the guava cache has finished loading the key to avoid a race condition where the watcher
+        // causes us to invalidate an entry before the key has been fully loaded.
+        final WatcherHandle<FileWatcher> watcherHandle = resourceWatcherService.add(fileWatcher);
+
+        return new KNNIndexCacheEntry(knnIndex, watcherHandle);
     }
-}
 
+    private static class KNNIndexCacheEntry {
```
vamshin, line 214: Minor: java doc for class

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -103,10 +106,12 @@ public synchronized void rebuild() {
      *
      * @param removalNotification key, value that got evicted.
      */
-    private void onRemoval(RemovalNotification<String, KNNIndex> removalNotification) {
-        KNNIndex knnIndex = removalNotification.getValue();
+    private void onRemoval(RemovalNotification<String, KNNIndexCacheEntry> removalNotification) {
+        KNNIndexCacheEntry knnIndexCacheEntry = removalNotification.getValue();
+
+        knnIndexCacheEntry.getFileWatcherHandle().stop();
```
vamshin, line 112: Should we add try/finally block and move ```knnIndexCacheEntry.getFileWatcherHandle().stop();``` to finally block to ensure the watcher is removed as file is deleted at this point?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -175,18 +178,57 @@ public void setCacheCapacityReached(Boolean value) {
...
-        return KNNIndex.loadIndex(indexPathUrl, algoParams);
+        FileWatcher fileWatcher = new FileWatcher(indexPath);
+        fileWatcher.addListener(KNN_INDEX_FILE_DELETED_LISTENER);
+
+        // Calling init() on the FileWatcher will bootstrap initial state that indicates whether or not the file
+        // is present. If it is not present at time of init(), then KNNIndex.loadIndex will fail and we won't cache
+        // the entry
+        fileWatcher.init();
+
+        final KNNIndex knnIndex = KNNIndex.loadIndex(indexPathUrl, algoParams);
```
vamshin, line 199: I think we should first do  ```resourceWatcherService.add``` and then load the index.
- ```resourceWatcherService.add``` seems to already call ```init()``` on the filewatcher.  So no explicit call to init()
-  this will also address the case of  index not loaded if file is only present

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -175,18 +178,57 @@ public void setCacheCapacityReached(Boolean value) {
...
+
+        // TODO verify that this is safe - ideally we'd explicitly ensure that the FileWatcher is only checked
+        // after the guava cache has finished loading the key to avoid a race condition where the watcher
+        // causes us to invalidate an entry before the key has been fully loaded.
+        final WatcherHandle<FileWatcher> watcherHandle = resourceWatcherService.add(fileWatcher);
```
vamshin, line 204: I think guava cache takes the lock when loading the entry for the key. So when invalidate happens, it should happen after the graph is loaded? 
 https://github.com/google/guava/blob/master/guava/src/com/google/common/cache/Cache.java#L91

vamshin (commented):
Good Catch. thanks for the PR.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
jschmitz28: Will add

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
jschmitz28: Can you clarify what would be wrapped in the try{}? ```knnIndexCacheEntry.getFileWatcherHandle().stop()``` should not throw - it just removes the watcher from the set, and doesn't declare any kind of exceptions.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
jschmitz28: See the following gist. Guava invalidate(key) does not block but is instead a no-op when executed if guava hasn't fully loaded the key into the cache: https://gist.github.com/jschmitz28/c8ff2d3b5856f625cf1f5a98095b7a51

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
vamshin: I was referring to ensuring ```knnIndexCacheEntry.getFileWatcherHandle().stop()``` being called when cache invalidate happens. some thing like this

```
try {
        KNNIndexCacheEntry knnIndexCacheEntry = removalNotification.getValue();
               executor.execute(() -> knnIndexCacheEntry.getKnnIndex().close());
} finally {
       knnIndexCacheEntry.getFileWatcherHandle().stop();
}

```

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
jschmitz28: Can you clarify what you mean by the second case?

For the first case, the reason why I'm calling ```loadIndex``` in between ```FileWatcher.init``` and ```resourceWatcherService.add``` is because I'm still not totally confident about the guarantee of no search threads executing while the file is deleted.

By calling ```init()``` first, we know with certainty whether the file existed at the time we started loading the graph, and if it did not exist, we are guaranteed that ```loadIndex``` will fail. If we also start the monitor before loading the index, then in the case that we're wrong about the guarantee mentioned above, we're effectively creating a multi-second window (depending on index loading time) for which the watcher could notify about a file delete, and invalidate() would cause a no-op.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
jschmitz28: Is there a perceived difference in behavior here? Both should be called regardless - in the case of the PR, ```FileWatcher.stop``` is called first, and should doesn't have an expected failure mode, and then we submit a task to ```gc``` the index.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
vamshin: makes sense. Thanks

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
vamshin: Got it. You mean invalidating entry would not interrupt loading graph but no-op would make us miss ```gc()``` on the graph.  Agree this minimizes the risk of race condition assuming the index deletion happens during search.

vamshin (approved):
LGTM! Thanks for the changes

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
vamshin: Agree there is no failure mode at this point. Ideally we want to ensure both gc() and stop() always get called in this function. We can defer this until we use functions that throws.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -121,15 +126,13 @@ private void onRemoval(RemovalNotification<String, KNNIndex> removalNotification
...
-     * a failure to load the
-     * @throws RuntimeException if there's an unexpected failure in loading, which implies that the value for
-     * the key will be both out of the cache and the underlying index will not be loaded
+     * @param algoParams hnsw algorithm parameters
+     * @return KNNIndex holding the heap pointer of the loaded graph
      */
     public KNNIndex getIndex(String key, final String[] algoParams) {
         try {
-            return cache.get(key, () -> loadIndex(key, algoParams));
+            final KNNIndexCacheEntry knnIndexCacheEntry = cache.get(key, () -> loadIndex(key, algoParams));
```
jmazanec15, line 134: Why not just do:
```
return cache.get(key, () -> loadIndex(key, algoParams)).getKnnIndex();
```

jmazanec15 (approved):
LGTM thanks for the fix!

PR #38, opened 2020-02-04
Add index wrapper

jschmitz28:
### Description
Add index wrapper  o avoid manual memory management and re-creationf space object on all queries, and simplify error handling logic a bit

Note that I think that it also makes sense to declare a "throws IOException" in KnnIndex.java but haven't gotten around to it, posting this now to hopefully unblock you guys a bit


jschmitz28:
New PR branched off of new master here: https://github.com/opendistro-for-elasticsearch/k-NN/pull/39

PR #39, opened 2020-02-05
Add IndexWrapper to JNI code to reduce manual memory management and recreation of space on queries

jschmitz28:
Note: Supercedes https://github.com/opendistro-for-elasticsearch/k-NN/pull/38

### Description
This change adds IndexWrapper into the JNI layer similar to how nmslib does it in their python bindings. This avoids the need to recreate the space on all queries as well as provides automatic deletion upon deletion of the IndexWrapper.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/v1736/KNNIndex.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/v1736/KNNIndex.java
@@ -87,6 +87,12 @@ public class KNNIndex implements AutoCloseable {
     public void close() {
         Lock writeLock = readWriteLock.writeLock();
         writeLock.lock();
+
+        // Autocloseable documentation recommends making close idempotent. We don't expect to doubly close
+        // but this will help prevent a crash in that situation.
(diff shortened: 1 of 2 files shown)


jschmitz28:
Note that so far I don't have a working setup to run end to end integration tests with a custom JNI layer, all that's tested so far in this is compilation

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/v1736/KNNIndex.java
```diff
@@ -87,6 +87,12 @@ public long getIndexSize() {
     public void close() {
         Lock writeLock = readWriteLock.writeLock();
         writeLock.lock();
+
+        // Autocloseable documentation recommends making close idempotent. We don't expect to doubly close
+        // but this will help prevent a crash in that situation.
+        if (this.isClosed) {
+            return;
+        }
```
vamshin, line 95: +1

vamshin (approved):
LGTM!

jmazanec15 (approved):
Looks good to me! Thanks for the contribution

PR #44, opened 2020-02-08
updated readme

jmazanec15:
*Issue #, if available:*
#40 

*Description of changes:*
Added description of settings and stats to the Readme. Additionally, I added Credits and Acks section.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,10 +2,6 @@
 
 Open Distro for Elasticsearch enables you to run nearest neighbor search on billions of documents across thousands of dimensions with the same ease as running any regular Elasticsearch query. You can use aggregations and filter clauses to further refine your similarity search operations. K-NN similarity search power use cases such as product recommendations, fraud detection, image and video search, related document search, and more.
 
-## REQUEST FOR COMMENT (RFC)
-
-We'd like to get your comments! Please read the plugin RFC [document](https://github.com/opendistro-for-elasticsearch/k-NN/blob/development/RFC.md) and raise an issue to add your comments and questions.
-
 ## Documentation
 
 To learn more, please see our [documentation](https://opendistro.github.io/for-elasticsearch-docs/).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,89 +53,223 @@ The following code creates a KNN index with fields my_vector1, my_vector2, my_ve
...
+        }
+    },
+    "mappings": {
+        "properties": {
+            "my_vector": {
+                "type": "knn_vector",
+                "dimension": 2
+            }
+        }
     }
```
vamshin, line 76: Should we have one example showing index creation with multiple knn_vector fields?

README.md
```diff
@@ -57,89 +53,223 @@ The following code creates a KNN index with fields my_vector1, my_vector2, my_ve
...
 }
-'
+```
+
+## Java Native library usage
+For plugin installations from archive(.zip), it is necessary to ensure ```.so``` file for linux OS and ```.jnilib``` file for Mac OS are present in the java library path. This can be possible by copying .so/.jnilib to either $ES_HOME or by adding manually ```-Djava.library.path=<path_to_lib_files>``` in ```jvm.options``` file
+
+## Settings
+### Index Level Settings
+KNN Index level settings should be provided at the time of the index creation. If the settings are not provided, default values will be use. These settings are static which means they cannot be changed after index creation.
```
aetter: You must provide index-level settings when you create the index. If you don't provide these settings, KNN uses its default values. These settings are static, which means you can't modify them after index creation.

README.md
```diff
@@ -57,89 +53,223 @@ The following code creates a KNN index with fields my_vector1, my_vector2, my_ve
...
+
+##### index.knn.algo_param.ef_search
+This setting is an HNSW parameter that represents "the size of the dynamic list for the nearest neighbors (used during the search). Higher ef leads to more accurate but slower search." [nmslib/hnswlib](https://github.com/nmslib/hnswlib/blob/master/ALGO_PARAMS.md) The default value is *512*.
+
+##### index.knn.algo_param.ef_construction
+This setting is an HNSW parameter that "the parameter has the same meaning as ef, but controls the index_time/index_accuracy. Bigger ef_construction leads to longer construction, but better index quality." [nmslib/hnswlib](https://github.com/nmslib/hnswlib/blob/master/ALGO_PARAMS.md) The default value is *512*.
+
+##### More Information
```
aetter: Probably OK to remove this section, given that we just linked to the page three times.

README.md
```diff
@@ -57,89 +53,223 @@ The following code creates a KNN index with fields my_vector1, my_vector2, my_ve
...
+### Cluster Stats
+#### circuit_breaker_triggered
+Indicates whether the circuit breaker is triggered.
+
+### Node Stats
+#### eviction_count
+The number of evictions that have occurred in the guava cache. *note:* explicit evictions that occur because of index deletion are not counted.
+
+#### hit_count
+The number of cache hits that have occurred on the node.
```
aetter: Can you clarify the meaning here? As in, a user ran a search, and the particular graph the user was trying to search was in memory, so that's... one hit? A hit for each result returned?

README.md
```diff
@@ -57,89 +53,223 @@ The following code creates a KNN index with fields my_vector1, my_vector2, my_ve
...
+
+#### miss_count
+The number of cache misses that have occurred on the node.
```
aetter: Ditto here. A user ran a search, the particular graph wasn't in memory, so KNN had to load it in...? Does that constitute one miss?

README.md
```diff
@@ -57,89 +53,223 @@ The following code creates a KNN index with fields my_vector1, my_vector2, my_ve
...
+        "knn.circuit_breaker.unset.percentage": 23
+    }
 }
-'
+```
+
+## Stats
+The KNN Stats API provides information about the current status of the KNN Plugin. The plugin keeps track of both cluster level and node level stats. Cluster level stats have a single value for the entire cluster. Node level stats have a single value for each node in the cluster. A user can filter their query by nodeID and statName in the following way:
+```
+GET /_opendistro/_knn/nodeId1,nodeId2/stats/statName1,statName2
```
aetter, line 193: Another excellent example.

README.md
```diff
@@ -57,89 +53,223 @@ The following code creates a KNN index with fields my_vector1, my_vector2, my_ve
...
+#### Circuit Breaker
+For KNN, the circuit breaker is used to indicate when performance may degrade because the graphs loaded into native memory are reaching the cluster’s total limits. Currently, the system does not perform any action once this limit is reached.
+
+##### knn.memory.circuit_breaker.enabled
+This setting enables or disables the circuit breaker feature.  Disabling this setting will keep you at risk of Out of memory as we do not have control on the memory usage for the graphs. The default value is *true*. 
+
+##### knn.memory.circuit_breaker.limit
+This setting indicates the maximum capacity of the cache. When the cache attempts to load in a graph that exceeds this limit, it is forced to evict an entry and *knn.circuit_breaker.triggered *is set to *true.* The default value for this setting is *60% *of the machines total memory outside the Elasticsearch jvm . However, a value in *KB* can be given as well.
 
+###### Example
```
aetter, line 165: Great example.

aetter (approved):

README.md
```diff
@@ -57,89 +53,223 @@ The following code creates a KNN index with fields my_vector1, my_vector2, my_ve
 ```
 PUT /myindex
 {
-  "settings" : {
-    "index": {
-      "knn": true
+    "settings" : {
+        "number_of_shards" :   1,
+        "number_of_replicas" : 0,
```
vamshin: we can remove these settings?

README.md
```diff
@@ -57,89 +53,223 @@ The following code creates a KNN index with fields my_vector1, my_vector2, my_ve
...
+##### Example
+```
+PUT /my_index/_settings
 {
-"my_vector" : [3.5, 4.5],
-"price":15
+    "index" : {
+        "index.knn.algo_param.m": 18, 
+        "index.knn.algo_param.ef_search" : 20,
+        "index.knn.algo_param.ef_construction" : 40,
```
vamshin: we should remove ```index``` from setting as we already wrapped settings in ```index```?

It should be 

```
"index" : {
        "knn.algo_param.m": 18, 
        "knn.algo_param.ef_search" : 20,
        "knn.algo_param.ef_construction" : 40
}

```

jmazanec15: Will update thanks

README.md
jmazanec15: Will remove thanks

README.md
jmazanec15: Will update thanks

README.md
jmazanec15: Makes sense thanks

README.md
jmazanec15: Yes will clarify

README.md
jmazanec15: Will clarify

README.md
jmazanec15: Will update thanks

README.md
```diff
@@ -57,89 +53,226 @@ The following code creates a KNN index with fields my_vector1, my_vector2, my_ve
...
+This setting is an HNSW parameter that "the parameter has the same meaning as ef, but controls the index_time/index_accuracy. Bigger ef_construction leads to longer construction, but better index quality." [nmslib/hnswlib](https://github.com/nmslib/hnswlib/blob/master/ALGO_PARAMS.md) The default value is *512*.
+
+##### Example
+```
+PUT /my_index/_settings
 {
-"my_vector" : [3.5, 4.5],
-"price":15
+    "index" : {
+        "knn.algo_param.m": 18,
```
vamshin, line 132: should we also include ```"knn":true```?

jmazanec15: Yes will update

vamshin (approved):
LGTM! Thanks for enriching ```README```.

Issue #46, opened 2020-02-09
Add gradle task to run multinode cluster locally


jmazanec15:
RestKNNStatsHandlerIT should run with multiple node cluster to check for issues with Transport Layer.

Issue #45, opened 2020-02-08
Stats Issue for Transport Layer for ES 7.4


jmazanec15:
```
GET /_opendistro/_knn/stats?pretty
      {
        "type" : "failed_node_exception",
        "reason" : "Failed node [Xc40NDAATJ2t-oYTUUztYw]",
        "node_id" : "Xc40NDAATJ2t-oYTUUztYw",
        "caused_by" : {
          "type" : "illegal_state_exception",
          "reason" : "Message not fully read (request) for requestId [213900], action [cluster:admin/knn_stats_action[n]], available [3180]; resetting"
        }
      },
      {
        "type" : "failed_node_exception",
        "reason" : "Failed node [DQL5evtaTZ-etbLHoVMQsw]",
        "node_id" : "DQL5evtaTZ-etbLHoVMQsw",
        "caused_by" : {
          "type" : "illegal_state_exception",
          "reason" : "Message not fully read (request) for requestId [213901], action [cluster:admin/knn_stats_action[n]], available [3180]; resetting"
        }
      }
    ]
  },
  "circuit_breaker_triggered" : false,
  "nodes" : {
    "0TBag8zVSt2AP050KrLVYw" : {
      "eviction_count" : 0,
      "miss_count" : 0,
      "graph_memory_usage" : 0,
      "cache_capacity_reached" : false,
      "load_exception_count" : 0,
      "hit_count" : 0,
      "load_success_count" : 0,
      "total_load_time" : 0
    }
  }
}
```

PR #47, opened 2020-02-09
added reading from stream input for KNN stats request

jmazanec15:
*Issue #, if available:*
#45 
*Description of changes:*
Minor fix to read from StreamInput for KNNStatsRequest in KNNStatsNodeRequest. 

Verified fix worked by spinning up 2 node cluster and checking stats output. 

RestKNNStatsHandlerIT should run on multiple node cluster instead of single node to catch issues related to Transport Layer. #46 


vamshin (approved):
Good catch. Thanks for  fixing @jmazanec15 .

Issue #51, opened 2020-02-16
Lazy load efSearch parameter.


vamshin:
efSearch Parameter is required only during the graph loading.  
Read efSearch from Index settings when loading graph for the first time.

PR #52, opened 2020-02-16
lazy load efSearch

vamshin:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/k-NN/issues/51

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
@@ -126,12 +126,12 @@ public class KNNIndexCache {
      * Loads corresponding index for the given key to memory and returns the index object.
      *
      * @param key indexPath where the serialized hnsw graph is stored
-     * @param algoParams hnsw algorithm parameters
+     * @param indexName index name
      * @return KNNIndex holding the heap pointer of the loaded graph
      */
-    public KNNIndex getIndex(String key, final String[] algoParams) {
+    public KNNIndex getIndex(String key, final String indexName) {
         try {
-            final KNNIndexCacheEntry knnIndexCacheEntry = cache.get(key, () -> loadIndex(key, algoParams));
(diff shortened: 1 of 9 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -236,4 +236,8 @@ public void onFileDeleted(Path indexFilePath) {
             getInstance().cache.invalidate(indexFilePath.toString());
         }
     };
+
+    private String[] getQueryParams(String indexName) {
+        return new String[] {"efSearch=" + KNNSettings.getEfSearchParam(indexName)};
```
jmazanec15, line 241: Could you explain more why there is only 1 query parameter being passed in? Why are M and efConstruction not gotten?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/v1736/KNNIndex.java
```diff
@@ -87,12 +87,6 @@ public long getIndexSize() {
     public void close() {
         Lock writeLock = readWriteLock.writeLock();
         writeLock.lock();
-
```
jmazanec15, line 90: Why is this no longer needed?

jmazanec15 (commented):
Just have a couple questions. Overall looks good to me

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/v1736/KNNIndex.java
vamshin: Good catch. Bad merge. Will add it back.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
vamshin: ```efSearch``` is required only during searches.  Other algorithm params are needed for indexing.

jmazanec15 (approved):
LGTM thanks!

Issue #62, opened 2020-03-09
Issue in CB with Stats call


jmazanec15:
The stats call in the CB [runnable](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNCircuitBreaker.java#L82) uses a constructor for [KNNStatsRequest](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/transport/KNNStatsRequest.java#L41) that is only supposed to be used for the transport action. This leads to null pointer exception.

PR #63, opened 2020-03-09
fixed cb

jmazanec15:
*Issue #, if available:*
#62 
*Description of changes:*
Fix KNNStatsRequest call in Circuit Breaker and add an Integ Test case to validate CB.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNCircuitBreaker.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNCircuitBreaker.java
@@ -15,6 +15,7 @@
 
 package com.amazon.opendistroforelasticsearch.knn.index;
 
+import com.amazon.opendistroforelasticsearch.knn.plugin.stats.KNNStatsConfig;
 import com.amazon.opendistroforelasticsearch.knn.plugin.stats.StatNames;
 import com.amazon.opendistroforelasticsearch.knn.plugin.transport.KNNStatsAction;
 import com.amazon.opendistroforelasticsearch.knn.plugin.transport.KNNStatsNodeResponse;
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNCircuitBreaker.java
```diff
@@ -35,6 +36,7 @@
  */
 public class KNNCircuitBreaker {
     private static Logger logger = LogManager.getLogger(KNNCircuitBreaker.class);
+    public static int CB_TIME_INTERVAL = 2*60;
```
vamshin, line 40: Nit: Add comment to mention seconds?

jmazanec15: Will add

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNCircuitBreakerIT.java
```diff
@@ -0,0 +1,98 @@
...
+
+    public void testCbTripped() throws Exception {
+        tripCb();
+    }
+
+    public void testCbUntrips() throws Exception {
+        updateClusterSettings("knn.circuit_breaker.triggered", "true");
+        //TODO: Attempt to find a better way to trigger runnable than waiting 2 minutes for it to finish
+        Thread.sleep(CB_TIME_INTERVAL*1000);
+        assertFalse(isCbTripped());
```
vamshin: Do we still need to sleep here?

vamshin: Looks like BackoffPolicy takes care of waiting. We can remove these assertions?

jmazanec15: Good catch. I must have forgotten to remove that

jmazanec15: Yes. Will update.

vamshin (approved):
LGTM!

Issue #65, opened 2020-03-11
KNN80HnswIndexTests.testFooter fails rarely


jmazanec15:
KNN80HnswIndexTests.testFooter fails very infrequently with the following error:

```
com.amazon.opendistroforelasticsearch.knn.index.KNN80HnswIndexTests > testFooter FAILED
    java.lang.ClassCastException: class org.apache.lucene.index.ParallelCompositeReader$1 cannot be cast to class org.apache.lucene.index.SegmentReader (org.apache.lucene.index.ParallelCompositeReader$1 and org.apache.lucene.index.SegmentReader are in unnamed module of loader 'app')
        at __randomizedtesting.SeedInfo.seed([DDF9C7E6F06832C8:B63D1282907F2A19]:0)
        at com.amazon.opendistroforelasticsearch.knn.index.KNNWeight.scorer(KNNWeight.java:72)
        at org.apache.lucene.search.Weight.bulkScorer(Weight.java:181)
        at org.apache.lucene.search.LRUQueryCache$CachingWrapperWeight.bulkScorer(LRUQueryCache.java:825)
        at org.apache.lucene.search.AssertingWeight.bulkScorer(AssertingWeight.java:88)
        at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:648)
        at org.apache.lucene.search.AssertingIndexSearcher.search(AssertingIndexSearcher.java:72)
        at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:440)
        at org.apache.lucene.search.IndexSearcher.search(IndexSearcher.java:564)
        at org.apache.lucene.search.IndexSearcher.count(IndexSearcher.java:366)
        at com.amazon.opendistroforelasticsearch.knn.index.KNN80HnswIndexTests.testFooter(KNN80HnswIndexTests.java:87)
REPRODUCE WITH: ./gradlew ':test' --tests "com.amazon.opendistroforelasticsearch.knn.index.KNN80HnswIndexTests.testFooter" -Dtests.seed=DDF9C7E6F06832C8 -Dtests.security.manager=false -Dtests.locale=ar-MA -Dtests.timezone=Europe/Zagreb -Dcompiler.java=12 -Druntime.java=12


```

Here is the command with the seed that causes the failure: `./gradlew ':test' --tests "com.amazon.opendistroforelasticsearch.knn.index.KNN80HnswIndexTests.testFooter" -Dtests.seed=DDF9C7E6F06832C8`

PR #66, opened 2020-03-12
fixed issue with test footer test case

jmazanec15:
*Issue #, if available:*
#65 
*Description of changes:*
Fixed rare test failure in testFooter. To get the `indexSearcher`, we were using a method from Lucene Test Case called `newSearcher`. `newSearcher` uses randomization when testing. `newSearcher` calls `maybeWrapReader`, which "rarely" calls `wrapReader`. In `wrapReader`, it sometimes wraps the reader with `ParallelLeafReader`, which is not ES compatible.

So, the solution is to call the IndexSearcher constructor instead of `newSearcher.`
`


vamshin (approved):
LGTM! Good catch.

Issue #71, opened 2020-03-20
Upgrade to ODFE 1.6


jmazanec15:
Upgrade KNN to ODFE 1.6. This will require upgrade to ES version 7.6. Additionally, this requires changes to the KNN Codec and test structure.

PR #72, opened 2020-03-20
Upgrade 76

jmazanec15:
*Issue #, if available:*
#71 

*Description of changes:*
Upgraded to ODFE 1.6
* Update ODFE version
* Upgrade to ES 7.6
* Switch to Gradle 6.2.1
* Switch to Java 13 
* Modify 
* Create KNN84Codec

Additionally, cleaned up a few things
* added style error check and fixed style errors
* refactored tests


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,21 +1,5 @@
...
-#
-#       http://www.apache.org/licenses/LICENSE-2.0
-#
-#   or in the "license" file accompanying this file. This file is distributed
-#   on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either
-#   express or implied. See the License for the specific language governing
-#   permissions and limitations under the License.
-#
-
 distributionBase=GRADLE_USER_HOME
```
vamshin, line 16: Why did we remove Copyright info?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN80Codec/KNN80Codec.java
```diff
@@ -50,11 +43,11 @@
     private final CompoundFormat compoundFormat;
     private Codec lucene80Codec;
 
-    public static final String CODEC_NAME = "KNN80Codec";
-    public static final String LUCENE_CODEC = "Lucene80";
+    public static final String KNN_80 = "KNN80";
```
vamshin: We should  not be changing codec name. This can result in compatibility issues when migrated from ODFE1.4 to 1.6 as segments would be created with KNN80Codec and it would not be able to locate it.

gradle/wrapper/gradle-wrapper.properties
jmazanec15: Good catch. Will add back in

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN80Codec/KNN80Codec.java
jmazanec15: Makes sense. Will Revert everything back to KNNXXCodec naming convention

vamshin (approved):
LGTM!

PR #75, opened 2020-03-24
fix npe

jmazanec15:
*Issue #, if available:*
#71 

*Description of changes:*
When running against a remote cluster, the current tests would throw an NPE because of a null String. This PR fixes that.

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNRestTestCase.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNRestTestCase.java
@@ -73,7 +73,7 @@ public class KNNRestTestCase extends ESRestTestCase {
         // jacoco.dir is set in esplugin-coverage.gradle, if it doesn't exist we don't
         // want to collect coverage so we can return early
         String jacocoBuildPath = System.getProperty("jacoco.dir");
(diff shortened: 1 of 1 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNRestTestCase.java
```diff
@@ -73,7 +73,7 @@ public static void dumpCoverage() throws IOException, MalformedObjectNameExcepti
         // jacoco.dir is set in esplugin-coverage.gradle, if it doesn't exist we don't
         // want to collect coverage so we can return early
         String jacocoBuildPath = System.getProperty("jacoco.dir");
-        if (jacocoBuildPath.isEmpty()) {
+        if (jacocoBuildPath == null || jacocoBuildPath.isEmpty()) {
```
vamshin: We can use ```Strings.isNullOrEmpty``` from ```import org.elasticsearch.common.Strings;```

jmazanec15: Good catch. Will update

vamshin (approved):
LGTM!

Issue #77, opened 2020-03-25
Debugging functionality is broken ES 7.6


jmazanec15:
In order to debug integration tests, we attach a debugger to the ES Test Cluster like this: `./gradlew integTests --debug-jvm`. However, this does not work for ES 7.6.

PR #78, opened 2020-03-25
fixed debugging functionality

jmazanec15:
*Issue #, if available:*
#77 
*Description of changes:*
This PR fixes a bug preventing a debugger from being able to attach to the integration test server and updates the documentation. In prior Elasticsearch versions, it was possible to run the command `./gradlew integTests --debug-jvm` to connect the debugger to the integTests server. For ES 7.6, this is no longer possible. Now, we pass the JVM args to the integration tests server manually.

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,10 +22,10 @@ The package uses the [Gradle](https://docs.gradle.org/5.5.1/userguide/userguide.
 
 ### Debugging
 
-Sometimes it is useful to attach a debugger to either the Elasticsearch cluster or the integ tests to see what's going on. When running unit tests, hit **Debug** from the IDE's gutter to debug the tests.  For the Elasticsearch cluster or the integ tests, first, make sure that the debugger is listening on port `5005`. Then, to debug the server  code, run:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -34,15 +34,15 @@ OR
 ./gradlew run --debug-jvm # to just start a cluster that can be debugged
 ```
 
-The Elasticsearch server JVM will launch suspended and wait for a debugger to attach to `localhost:8000` before starting the Elasticsearch server.
+The Elasticsearch server JVM will connect to a debugger attached to `localhost:5005` before starting.
```
vamshin, line 37: why are we changing the port for ```run```?  Can we have ```run --debug-jvm``` on port 8000 and ``` :integTest  -Ddebug.es=1 ``` on port 5005 ?

jmazanec15: Why do we want to have them to run on 2 different ports? Is there a case where they would need to be run at the same time?

vamshin: Yes we may need to run both simultaneously in the case when you want to debug integ tests. Because integration tests run on different jvm and cluster on different jvm 

```
./gradlew run --debug-jvm -Ddebug.es=1
```

jmazanec15: Not sure why you would do `./gradlew run --debug-jvm -Ddebug.es=1`. Do you mean running something like:
`./gradlew integTest -Ddebug.es=1 -Dtest.debug=1`?

Also I will change `debug.es` to be consistent with naming conventions.

vamshin: We might need this when we want to debug integ test all the way to the Elasticsearch code.

vamshin (approved):
LGTM!

Issue #67, opened 2020-03-12
KNN80HnswIndexTests.testMultiFieldsKnnIndex


jmazanec15:
`testMultiFieldsKnnIndex` fails rarely due.

```
com.amazon.opendistroforelasticsearch.knn.index.KNN80HnswIndexTests > testMultiFieldsKnnIndex FAILED
    java.lang.AssertionError: expected:<4> but was:<2>
        at __randomizedtesting.SeedInfo.seed([A01C285FBB136CD0:1A5C3E6AD0FF55A3]:0)
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at com.amazon.opendistroforelasticsearch.knn.index.KNN80HnswIndexTests.testMultiFieldsKnnIndex(KNN80HnswIndexTests.java:129)
REPRODUCE WITH: ./gradlew ':test' --tests "com.amazon.opendistroforelasticsearch.knn.index.KNN80HnswIndexTests.testMultiFieldsKnnIndex" -Dtests.seed=A01C285FBB136CD0 -Dtests.security.manager=false -Dtests.locale=mzn -Dtests.timezone=Australia/Queensland -Dcompiler.java=12 -Druntime.java=12

```
Reproduce by running: `./gradlew ':test' --tests "com.amazon.opendistroforelasticsearch.knn.index.KNN80HnswIndexTests.testMultiFieldsKnnIndex" -Dtests.seed=A01C285FBB136CD0`

PR #86, opened 2020-04-06
Broken Test Cases

jmazanec15:
*Issue #, if available:*
#67 

*Description of changes:*
testMultiFieldsKnnIndex Test case infrequently creates 4 index files instead of 2. This is because the writer is not closed right after the document is added so segments do not get merged. To fix this, this PR closes the writer after the document is added.

Additionally, the 2 codec test classes were incorrectly named. This prevents them from running during `gradlew build`. So, this PR renames them as well.


vamshin (approved):
LGTM!

Issue #88, opened 2020-04-10
Add stats to track the number of requests and errors in KNN index/query operations


jmazanec15:
In order to better monitor KNN performance, it would be beneficial to add more book keeping about the total number of KNN index/query operations and errors on a node level that are exposed through the stats API.

PR #89, opened 2020-04-10
Add stats to track knn request counts

jmazanec15:
*Issue #, if available:*
#88 

*Description of changes:*
PR adds stats to track the number of requests and errors for KNN query and index operations.

For query operations, bookkeeping is added in the queryIndex function in KNNIndex. For index operations, it is added in the KNNVectorFieldMapper parse function.

Unit tests have been added to make sure that the counting functionality works properly.

Documentation has been updated as well.


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -230,46 +231,52 @@ public Query termQuery(Object value, QueryShardContext context) {
...
-        if (fieldType().stored()) {
-            context.doc().add(new StoredField(name(), point.toString()));
+            context.doc().add(point);
+            if (fieldType().stored()) {
+                context.doc().add(new StoredField(name(), point.toString()));
+            }
+            context.path().remove();
+        } catch (Exception ex) {
+            KNNCounter.GRAPH_INDEX_ERRORS.increment();
+            throw ex;
```
vamshin: These Exceptions will only include the parsing errors(customer error) which we could ignore. Graphs are indexed part of the KNN80DocValuesConsumer. We may want to keep track of failures for graph creation there

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/v1736/KNNIndex.java
```diff
@@ -84,6 +85,9 @@ public long getIndexSize() {
                     }
             );
 
+        } catch (Exception ex) {
+            KNNCounter.GRAPH_QUERY_ERRORS.increment();
+            throw new RuntimeException("Unable to query the index: " + ex);
```
vamshin, line 90: just a thought, why not we just rely on load_exception_count metric from cache stats. This seem to track number of exceptions while loading graph which will be invoked during queries?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -230,46 +231,52 @@ public Query termQuery(Object value, QueryShardContext context) {
 
     @Override
     public void parse(ParseContext context) throws IOException {
-        if (!KNNSettings.isKNNPluginEnabled()) {
-            throw new IllegalStateException("KNN plugin is disabled. To enable " +
-                                                    "update knn.plugin.enabled setting to true");
-        }
+        KNNCounter.GRAPH_INDEX_REQUESTS.increment();
```
vamshin: This might add a little confusion for bulk requests, which could index multiple vectors but still part of same request. If the intention is to count the number of graph requests , probably we could count at ```KNN80DocValuesConsumer```.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
jmazanec15: Oh okay that makes sense. Will update.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
jmazanec15: Yes, the intention of this metric is to count the total number of requests to index graphs. Will move to `KNN80DocValuesConsumer`

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/v1736/KNNIndex.java
jmazanec15: Using load_exception_count would only count exceptions for loading the graph into memory, not the actual query of the graph. Adding the metric here allows us to check if the library query of the graph fails. In your opinion, should this metric track the number of query errors where a query is a call to the ES search API for knn, or for a query where a query is a call to the library function?

README.md
```diff
@@ -215,6 +215,18 @@ The number of cache misses that have occurred on the node. A cache miss occurs w
...
+#### graph_index_requests
+The number of requests to add the knn_vector field of a document into a graph.
+
+#### graph_index_errors
+The number of requests to add the knn_vector field of a document into a graph that have produced an error.
+
+#### graph_query_requests
+The number of graph queries that have been made. 
+
+#### graph_query_errors
```
vamshin, line 253: How about we add metrics for counting KNNQueries?

jmazanec15: Makes sense. Will add

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/v1736/KNNIndex.java
vamshin: makes sense. It should track the number of query errors.

vamshin (approved):
LGTM! Thanks for adding the metrics.

Issue #28, opened 2020-02-03
Support cosine similarity[Experimental]


vamshin:
This sim is to help prioritize the cosine similarity feature. Please +1 if you end up here :)

mickdelaney:
We would be huge fans of this also. We currently do this out of ES

vamshin:
Merged to 1.7 release. Should be available part of 1.7 docker images.
We are marking this feature experimental as we are still evaluating the performance.

We will update this thread once we have done evaluation. Thank you

PR #90, opened 2020-04-12
FEAT: support cosine similarity

chenqi0805:
*Issue #, if available:*
#28 

*Description of changes:*
- Add "index.knn.space_type" as string param into KNNSettings and use validator for fast fail
- Pass in spaceType as String into jni functions (java/c++)
- Update vector field mapper to correctly parse the spaceType setting param
- Adapt KNNIndexCache
- Adapt existing test cases

*TODO*
- [x] check if "index.knn.space_type" should be reset or not once a index is created
- [x] More test cases to cover cosine similarity
- [x] update README


jni/src/v1736/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp
```diff
@@ -85,15 +85,17 @@ void catch_cpp_exception_and_throw_java(JNIEnv* env)
...
+JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex_saveIndex(JNIEnv* env, jclass cls, jintArray ids, jobjectArray vectors, jstring indexPath, jstring spaceType, jobjectArray algoParams)
 {
     Space<float>* space = NULL;
     ObjectVector dataset;
     Index<float>* index = NULL;
     int* object_ids = NULL;
 
     try {
-        space = SpaceFactoryRegistry<float>::Instance().CreateSpace("l2", AnyParams());
+        const char *spaceTypeCStr = env->GetStringUTFChars(spaceType, 0);
```
jmazanec15, line 96: I believe this needs to be released via `env->ReleaseStringUTFChars`

jni/src/v1736/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp
```diff
@@ -171,17 +173,19 @@ JNIEXPORT jobjectArray JNICALL Java_com_amazon_opendistroforelasticsearch_knn_in
...
 }
 
-JNIEXPORT jlong JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex_init(JNIEnv* env, jclass cls,  jstring indexPath, jobjectArray algoParams)
+JNIEXPORT jlong JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex_init(JNIEnv* env, jclass cls,  jstring indexPath, jstring spaceType, jobjectArray algoParams)
 {
     IndexWrapper *indexWrapper = NULL;
     try {
         const char *indexPathCStr = env->GetStringUTFChars(indexPath, 0);
         string indexPathString(indexPathCStr);
+        const char *spaceTypeCStr = env->GetStringUTFChars(spaceType, 0);
```
jmazanec15: This needs to be released as well

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNJNITestsIT.java
```diff
@@ -31,8 +31,8 @@
 import java.util.Map;
 import java.util.stream.Collectors;
 
-public class KNNJNITests extends ESTestCase {
-    private static final Logger logger = LogManager.getLogger(KNNJNITests.class);
+public class KNNJNITestsIT extends ESTestCase {
```
jmazanec15: Why switch this?

jmazanec15 (commented):
Thanks for working on this! Added a couple comments. Additionally, nmslib submodule should not be altered.

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNJNITestsIT.java
chenqi0805: Ah. When I ran the test locally on IntelliJ, it fails to detect the integration tests therein due to lack of suffix "IT" in the class name. So I switched that for easier debugging. Will change it back but I think it might worth doublecheck if CI also neglects the tests.

jni/src/v1736/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp
chenqi0805: Ah. Thanks for the catch. Recalled being told that memory leak can be disastrous

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
```diff
@@ -367,4 +387,48 @@ public void setClusterService(ClusterService clusterService) {
         this.clusterService = clusterService;
     }
 
+    static class SpaceTypeValidator implements Setting.Validator<String> {
+
+        private Set<String> types = SpaceTypes.getValues();
+
+        @Override public void validate(String value) {
+            if (!types.contains(value.toLowerCase())){
```
jmazanec15: Why is this case insensitive?

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNJNITests.java
```diff
@@ -165,7 +214,7 @@ public Void run() {
         float[] queryVector = {1.0f, 1.0f, 1.0f, 1.0f};
         String[] algoQueryParams = {"efSearch=200"};
 
-        final KNNIndex index = KNNIndex.loadIndex(indexPath, algoQueryParams);
+        final KNNIndex index = KNNIndex.loadIndex(indexPath, "l2", algoQueryParams);
         final KNNQueryResult[] results = index.queryIndex(queryVector, 30);
 
         Map<Integer, Float> scores = Arrays.stream(results).collect(
```
jmazanec15, line 220: Could you add test cases to ensure that invalid space names throw the expected exceptions?

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNJNITestsIT.java
jmazanec15: Tests as opposed to IT signals that it is a unit test. These tests run during `./gradlew build`, which is what we use for the PR testing [workflow](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/.github/workflows/test-workflow.yml#L26). You can also run them individually with this command:
```
./gradlew ':test' --tests "com.amazon.opendistroforelasticsearch.knn.index.KNNJNITests.testCreateHnswIndex"
```
We consider these unit tests because they are testing an isolated portion of code. Interesting that IntelliJ does not pick them up. Will need to look into it.

chenqi0805: Ah. Sorry, I misidentified it as IT.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
chenqi0805: I intended to set it case insensitive, e.g. L2, Cosinesimil is also valid string. But I can remove this if not necessary.

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNJNITests.java
chenqi0805: Yes. My initial thought was to add such tests at KNNSettings module b/c that is where I want the input value to be validated and fail fast. Do you think it is also necessary to prevent invalid spaceType into JNI functions?

chenqi0805: Added an integration test in KNNESSettingsIT.

chenqi0805:
@jmazanec15 I have added one more test coverage on invalid spaceType setting and also updated README. The only concern left is whether spaceType setting is allowed to be dynamic or not. I think this depends on whether the KNN index setting parameters are embedded into fields and docs when stored, which I am not sure. Do you have idea here? Besides that, this work is now complete.

vamshin:
> @jmazanec15 I have added one more test coverage on invalid spaceType setting and also updated README. The only concern left is whether spaceType setting is allowed to be dynamic or not. I think this depends on whether the KNN index setting parameters are embedded into fields and docs when stored, which I am not sure. Do you have idea here? Besides that, this work is now complete.

spaceType setting should be static for an index to lead consistent results otherwise segments in the shards would be scattered with different spaces and queries would lead inconsistent results

jni/src/v1736/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp
```diff
@@ -85,15 +85,19 @@ void catch_cpp_exception_and_throw_java(JNIEnv* env)
     }
 }
 
-JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex_saveIndex(JNIEnv* env, jclass cls, jintArray ids, jobjectArray vectors, jstring indexPath, jobjectArray algoParams)
+JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex_saveIndex(JNIEnv* env, jclass cls, jintArray ids, jobjectArray vectors, jstring indexPath, jstring spaceType, jobjectArray algoParams)
```
vamshin: nit: Please add the new function arguments to the end of the existing parameters list. Please take care in other places as well.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
```diff
@@ -367,4 +387,48 @@ public void setClusterService(ClusterService clusterService) {
         this.clusterService = clusterService;
     }
 
+    static class SpaceTypeValidator implements Setting.Validator<String> {
+
+        private Set<String> types = SpaceTypes.getValues();
+
+        @Override public void validate(String value) {
```
vamshin, line 393: null check on ```value```

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
```diff
@@ -367,4 +387,48 @@ public void setClusterService(ClusterService clusterService) {
...
+            if (!types.contains(value.toLowerCase())){
+                throw new InvalidParameterException(String.format("Unsupported space type: %s", value));
+            }
+        }
+    }
+
+    /**
+     * Enum contains space types for k-NN similarity search
+     */
+    public enum SpaceTypes {
```
vamshin: nit: Move this enum class to a dedicated file

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/v1736/KNNIndex.java
```diff
@@ -110,12 +110,13 @@ public void close() {
      * Loads the knn index to memory for querying the neighbours
      *
      * @param indexPath path where the hnsw index is stored
+     * @param spaceType space type of the index
      * @param algoParams hnsw algorithm parameters
      * @return knn index that can be queried for k nearest neighbours
      */
-    public static KNNIndex loadIndex(String indexPath, final String[] algoParams) {
+    public static KNNIndex loadIndex(String indexPath, String spaceType, final String[] algoParams) {
```
vamshin: nit: final String spaceType?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
jmazanec15: I think case insensitive is good

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNJNITests.java
jmazanec15: Makes sense thanks

jmazanec15 (approved):
LGTM! Thanks for this contribution!

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
chenqi0805: @vamshin I put a null check here, but it still fails to guard against null value, i.e. for some reason, if index.knn.space_type == null, it will not go through the validate. Similar bug happens to other parameters, e.g.

```
PUT /myindex
{
    "settings" : {
        "index": {
            "knn": true,
            "knn.algo_param.m": null
        }
    }
}
```
```
PUT /myindex/_doc/2?refresh=true
{
    "my_vector1" : [1.5, 2.5],
    "price":10
}
```
```
POST /myindex/_search
{
    "size" : 10,
    "query": {
        "knn": {
            "my_vector1": {
                "vector": [15, 25],
                "k": 2
            }
        }
    }
}
```
yields
```
{
  "took" : 0,
  "timed_out" : false,
  "_shards" : {
    "total" : 1,
    "successful" : 1,
    "skipped" : 0,
    "failed" : 0
  },
  "hits" : {
    "total" : {
      "value" : 0,
      "relation" : "eq"
    },
    "max_score" : null,
    "hits" : [ ]
  }
}
```

How about we open a separate issue to track this?

chenqi0805:
@vamshin Thanks for the comments. Besides the comment on null value check, it is good for another look.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
vamshin: Ah got it. I think we can probably ignore this case.

vamshin (approved):
LGTM! Thanks for the contribution. Much required feature for our community.  Adding new space types supported by nmslib should be much simpler going forward and we could support it as per the community request.

As this change involves JNI level code changes and new functionality, we would evaluate this commit with performance tests and recall tests and then consider for merge. We are in the process of building automated performance tests for the commits but till then we may need to do some manual evaluation.

We will update this thread with the results and you could merge it then. Thank you.

jmazanec15:
Hi @chenqi0805, I think we are going to check this in as an experimental feature, until we can test it more thoroughly. Could you add a section in the readme under [this section](https://github.com/opendistro-for-elasticsearch/k-NN#basic-usage) for Cosine Similarity usage with an example and mark it in the header as experimental? Like this: `Cosine Similarity Usage (experimental)`

chenqi0805:
@jmazanec15 Will do in free time.

chenqi0805:
@jmazanec15 updated.

jmazanec15:
Great thank you @chenqi0805! This is a great feature. Going to check it in.

PR #91, opened 2020-04-13
Add gradle functionality to create multi-node cluster

jmazanec15:
*Issue #, if available:*
#46 

*Description of changes:*
Added logic in build.gradle to run multi-node cluster locally for testing and integration tests. Verified that `./gradlew integTests -PnumNodes=N` for N = 2 and 3 passes. Additionally, verified that a debugger can be attached to each node at the same time.

Along with this, I updated documentation.

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNCircuitBreakerIT.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNCircuitBreakerIT.java
@@ -41,10 +41,11 @@ public class KNNCircuitBreakerIT extends KNNRestTestCase {
         // Set circuit breaker limit to 1 KB
         updateClusterSettings("knn.memory.circuit_breaker.limit", "1kb");
 
-        // Create Single Shard Index so that all data is hosted on a single node
+        // Create index with 1 primary and numNodes-1 replicas so that the data will be on every node in the cluster
+        int numNodes = Integer.parseInt(System.getProperty("cluster.number_of_nodes"));
         Settings settings = Settings.builder()
                 .put("number_of_shards", 1)
-                .put("number_of_replicas", 2)
+                .put("number_of_replicas", numNodes - 1)
                 .put("index.knn", true)
                 .build();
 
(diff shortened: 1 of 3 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNCircuitBreakerIT.java
```diff
@@ -44,7 +44,7 @@ private void tripCb() throws Exception {
         // Create Single Shard Index so that all data is hosted on a single node
```
vamshin: update the comment?

jmazanec15: Will do. Good catch!!

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNCircuitBreakerIT.java
```diff
@@ -41,10 +41,11 @@ private void tripCb() throws Exception {
         // Set circuit breaker limit to 1 KB
         updateClusterSettings("knn.memory.circuit_breaker.limit", "1kb");
 
-        // Create Single Shard Index so that all data is hosted on a single node
+        // Create index with 1 primary and numNodes-1 replicas so that the data will be on every node in the cluster
+        int numNodes = Integer.parseInt(System.getProperty("cluster.number_of_nodes"));
         Settings settings = Settings.builder()
                 .put("number_of_shards", 1)
-                .put("number_of_replicas", 2)
+                .put("number_of_replicas", numNodes - 1)
```
vamshin, line 48: +1.

Are there any other tests with replicas that need to be taken care of?

jmazanec15: `numNodes - 1` and the 1 primary ensures there is one shard per node, correct?

This test fails with only one shard on a multi-node cluster because it requires a shard from 2 different indices to be searched on the same node to trigger the circuit breaker. With this fix, `./gradlew build` passes.

vamshin (approved):
LGTM!

Issue #69, opened 2020-03-16
Add stats information indicating indices in memory


jmazanec15:
Add stats information about which indices are in the cache on which nodes and how much memory they are occupying on those nodes.

PR #97, opened 2020-04-23
Added index cache information in stats

jmazanec15:
*Issue #, if available:*
#69 

*Description of changes:*
Increasing cache visibility by adding a node stat, `indices_in_cache`, that maps index names to the number of graphs they have in the cache and the total graph_memory_usage that they are taking up in the cache (in KB).

For example,
```
"indices_in_cache" : {
    "myindex" : {
        "graph_memory_usage" : 2,
        "graph_count" : 2
    }
},
```

In order to implement this, a few changes had to be made to the KNNIndexCache. 
1. In KNNIndexCacheEntry, the EsIndexName and the indexPathUrl were added as members
2. KNNIndexCache implements Closeable so that when the cache is finished being used in tests, the executor gets properly shutdown so that no threads leak
3. The following get functions were added: getWeightInKilobytes(String indexName), getGraphNamesForIndex(String indexName), getIndicesCacheStats()
4. 2 eviction methods were added as well: 1 to evict based on graph name (segment path), 1 evicts all entries from the cache. These were needed to test functionality related to the `indices_in_cache` stat, however, they will be needed for the clear cache API in the future.

In order to test, I created a `ESSingleNodeTestCase` test case for KNNIndexCache.


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -147,6 +162,35 @@ public CacheStats getStats() {
         return cache.stats();
     }
 
+    /**
+     * Get the Elasticsearch indices currently loaded into the Cache
+     *
+     * @return Set containing all of the Elasticsearch indices in the cache
```
vamshin: returns ```Map``` of index stats?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -174,6 +230,22 @@ public void setCacheCapacityReached(Boolean value) {
...
+     */
+    public void evictGraphFromCache(String indexFilePath) {
+        cache.invalidate(indexFilePath);
+    }
+
+    /**
+     * Evict all graphs in the cache manually
+     */
+    public void evictAllGraphsFromCache() {
+        cache.invalidateAll();
```
vamshin, line 268: Let us add info log here to help in debugging something like ```manually evicted all graphs from cache```

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -174,6 +230,22 @@ public void setCacheCapacityReached(Boolean value) {
         cacheCapacityReached.set(value);
     }
 
+    /**
+     * Evict a graph in the cache manually
+     *
+     * @param indexFilePath path to segment file. Also, key in cache
+     */
+    public void evictGraphFromCache(String indexFilePath) {
+        cache.invalidate(indexFilePath);
```
vamshin, line 260: lets add info log here ```Manually evicted graph from cache, indexFilePath, indexname```

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -156,6 +200,18 @@ public Long getWeightInKilobytes() {
         return cache.asMap().values().stream().map(KNNIndexCacheEntry::getKnnIndex).mapToLong(KNNIndex::getIndexSize).sum();
     }
 
+    /**
+     * Returns the current weight of an index in the cache in KiloBytes
+     *
+     * @param indexName Name if index to get the weight for
+     * @return Weight of the index in the cache in kilobytes
+     */
+    public Long getWeightInKilobytes(String indexName) {
```
vamshin: nit: ```final String indexName```

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -147,6 +162,35 @@ public CacheStats getStats() {
...
+     */
+    public Map<String, Map<String, Object>> getIndicesCacheStats() {
+        Map<String, Map<String, Object>> statValues = new HashMap<>();
+        String indexName;
+        for (Map.Entry<String, KNNIndexCacheEntry> index : cache.asMap().entrySet()) {
+            indexName = index.getValue().getEsIndexName();
+            statValues.putIfAbsent(indexName, new HashMap<>());
+
+            statValues.get(indexName).putIfAbsent(GRAPH_COUNT, 0);
+            statValues.get(indexName).put(GRAPH_COUNT, ((Integer) statValues.get(indexName).get(GRAPH_COUNT)) + 1);
```
vamshin: statValues.get(indexName).put(GRAPH_COUNT, ((Integer) statValues.get(indexName).getorDefault(GRAPH_COUNT, 0)) + 1)

vamshin (commented):
Thanks for enriching the stats. Few other stats which might be helpful as well

- graph_memory_usage_percentage in terms of total cache capacity at node level
- index_graph_memory_usage_percentage at index level for that node in terms of total cache capacity
- index_graph_loads at index level which counts the number of graph loads for an index in that node. This helps in identifying the hot indices in the cache. We could add a counter variable in the KNNIndexCacheEntry and increment it when ever the entry is accessed.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
jmazanec15: > Thanks for enriching the stats. Few other stats which might be helpful as well
> 
>     * graph_memory_usage_percentage in terms of total cache capacity at node level
> 
>     * index_graph_memory_usage_percentage at index level for that node in terms of total cache capacity
> 
>     * index_graph_loads at index level which counts the number of graph loads for an index in that node. This helps in identifying the hot indices in the cache. We could add a counter variable in the KNNIndexCacheEntry and increment it when ever the entry is accessed.

I think the first two make sense. I am not sure on the last one. Isn't the number of graphs per index somewhat analogous to index_graph_loads?

vamshin: Ah right. Third one seem to be already covered. You could ignore the third one.

vamshin (approved):

Issue #99, opened 2020-04-26
"NaN" vector fields are accepted. Docs with NaN fields should be rejected


vamshin:
NaN fields are accepted during indexing time but search is broken because of the bad data resulting in process crash.

### Reproduce Issue

> 

curl -X PUT "localhost:9200/myindex" -H 'Content-Type: application/json' -d'                                
{                       
  "settings" : {
  "number_of_shards" :   1,
  "number_of_replicas" : 0,
  "index": {                                     
      "knn": true
    }
  },
  "mappings": {
      "properties": {
        "my_vector": { 
          "type": "knn_vector",
          "dimension": 2
        }       
      }
  }
}
'

curl -X POST "localhost:9200/myindex/_doc/1?refresh=true" -H 'Content-Type: application/json' -d'                        
{                       
"my_vector" : ["NaN", "NaN"], 
"price":10
}                 
'

### Output
{"_index":"myindex","_type":"_doc","_id":"1","_version":2,"result":"updated","forced_refresh":true,"_shards":{"total":1,"successful":1,"failed":0},"_seq_no":1,"_primary_term":1}

> 

Code changes here https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java#L246

PR #100, opened 2020-04-27
Added validation that vector is not NaN

jmazanec15:
*Issue #, if available:*
#99 

*Description of changes:*
Added validation in VectorFieldMapper to check for vector values of NaN and throwing an Exception if so.

Added test case.


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -250,10 +250,16 @@ public void parse(ParseContext context) throws IOException {
         if (token == XContentParser.Token.START_ARRAY) {
             token = context.parser().nextToken();
             while (token != XContentParser.Token.END_ARRAY) {
+                if (Float.isNaN(context.parser().floatValue())) {
```
vamshin: Should we also add check for Float.isInfinite?

https://docs.oracle.com/javase/7/docs/api/java/lang/Float.html#isInfinite(float)

jmazanec15: Yes I think that makes sense.

vamshin (approved):
LGTM! Thanks for the fix

Issue #35, opened 2020-02-04
Create Gradle Task to Build KNN Library


jmazanec15:
Create a gradle task to build the KNN Library.

PR #105, opened 2020-05-06
Library build task

jmazanec15:
*Issue #, if available:*
#35 

*Description of changes:*
Adding task to build JNI library. Adding it as a dependency of test and integTest tasks so that library is built before tests are run.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,7 +48,9 @@ apply plugin: 'elasticsearch.esplugin'
 def usingRemoteCluster = System.properties.containsKey('tests.rest.cluster') || System.properties.containsKey('tests.cluster')
 def usingMultiNode = project.properties.containsKey('numNodes')
 // Only apply jacoco test coverage if we are running a local single node cluster
+def integTestDependOnJniLib = true
 if (!usingRemoteCluster && !usingMultiNode) {
+    integTestDependOnJniLib = false
     apply from: 'build-tools/knnplugin-coverage.gradle'
 }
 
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -109,13 +109,26 @@ thirdPartyAudit.enabled = false
...
+    workingDir 'jni'
+    commandLine 'cmake', '.'
+}
+
+task buildJniLib(type:Exec) {
+    dependsOn cmakeJniLib
+    workingDir 'jni'
+    commandLine 'make'
+}
+
```
vamshin, line 126: why not club this to one task

task buildJniLib(type:Exec) {
    workingDir 'jni'
    commandLine 'cmake', '.'
     commandLine 'make'
}

vamshin (commented):
Also can we add print statement in ```cmake ``` to output the location where we are dumping the shared library?

build.gradle
jmazanec15: The following syntax does not work. `commandLine 'cmake', '.'` is overridden with `commandLine 'make'`. 
```
task buildJniLib(type:Exec) {
workingDir 'jni'
commandLine 'cmake', '.'
commandLine 'make'
}
```

Additionally, having 2 separate tasks allows a user to just run the cmake task if they want.

jmazanec15:
> Also can we add print statement in `cmake ` to output the location where we are dumping the shared library?

Sure.

vamshin (approved):
LGTM! Building library part of gradlew build makes life easy for developers.

Issue #96, opened 2020-04-17
Block index writes to knn indices when the circuit breaker trips


vamshin:
As of today we do not block index writes to knn indices when graph memory circuit breaker triggers. We evict the entries in the cache to make room for new graphs.  One can keep adding more indices and result in the thrashing of cache which can degrade performance. We need to block the new index writes so that the performance does not degrade.

chenqi0805:
@vamshin Have taken a look at the issue and will work on it if there is no duplicate effort. My plan is to add a `KNNSettings.isCircuitBreakerTriggered()` check in `addKNNBinaryField`. Let me know if this sounds reasonable. 

By reading the design doc, the only other module I can think of adding this logic is `IndexWriter`. But then we need to inherit and override its api.

vamshin:
@chenqi0805 thanks for digging into the issue. You should be good to take this one.

We may need to reject the indexing requests at much earlier stages.  We could reject them even before parsing here https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java#L239.

PR #108, opened 2020-05-10
ENH: throws Exception in KNNVectorFieldMapper.parse API when circuit breaker trips

chenqi0805:
*Issue #, if available:*
#96 

*Description of changes:*
- Throws exception before parsing knn vector if the circuit breaker trips 
- Added two integration test cases

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
@@ -243,7 +243,8 @@ public class KNNVectorFieldMapper extends FieldMapper implements ArrayValueMappe
         }
 
         if (KNNSettings.isCircuitBreakerTriggered()) {
-            throw new IllegalStateException("Cannot update KNN vector field when circuit breaker is triggered.");
+            throw new IllegalStateException("Indexing knn vector fields is rejected as circuit breaker triggered." +
+                    " Check _opendistro/_knn/stats for detailed state");
         }
 
         context.path().add(simpleName());
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
@@ -242,6 +242,10 @@ public class KNNVectorFieldMapper extends FieldMapper implements ArrayValueMappe
                                                     "update knn.plugin.enabled setting to true");
         }
 
+        if (KNNSettings.isCircuitBreakerTriggered()) {
+            throw new IllegalStateException("Cannot update KNN vector field when circuit breaker is triggered.");
+        }
+
         context.path().add(simpleName());
 
         ArrayList<Float> vector = new ArrayList<>();
--- a/src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNESIT.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNESIT.java
@@ -16,12 +16,14 @@
 package com.amazon.opendistroforelasticsearch.knn.index;
 
 import org.apache.http.util.EntityUtils;
+import org.elasticsearch.client.Response;
 import org.elasticsearch.client.ResponseException;
 import org.elasticsearch.common.Strings;
 import org.elasticsearch.common.settings.Settings;
 import org.elasticsearch.common.xcontent.XContentFactory;
 
 import java.io.IOException;
+import java.util.List;
 
 import static org.hamcrest.Matchers.containsString;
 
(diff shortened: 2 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -242,6 +242,10 @@ public void parse(ParseContext context) throws IOException {
                                                     "update knn.plugin.enabled setting to true");
         }
 
+        if (KNNSettings.isCircuitBreakerTriggered()) {
```
jmazanec15, line 245: Should there also be a cluster setting like "CircuitBreakerOnTripEnableWriteBlock" so that users have the ability to opt out of the write block?

vamshin: IMO it would keep their performance at risk. Better to put write blocks on knn index and provide alert. They still have option to take risk by increasing `knn.memory.circuit_breaker.limit` percentage to a higher value

vamshin: > IMO it would keep their performance at risk. Better to put write blocks on knn index and provide alert. They still have option to take risk by increasing `knnmemorycircuit_breakerlimit` percentage

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -242,6 +242,10 @@ public void parse(ParseContext context) throws IOException {
...
+            throw new IllegalStateException("Cannot update KNN vector field when circuit breaker is triggered.");
```
vamshin: can we change message on the following lines "Indexing knn vector fields is rejected as circuit breaker triggered. Check _opendistro/_knn/stats for detailed state"

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNESIT.java
```diff
@@ -35,6 +35,20 @@ public void testAddKNNDoc() throws Exception {
...
+    public void testAddKNNDocBlockedWhenCbTrips() throws Exception {
+        createKnnIndex(INDEX_NAME, createKnnIndexMapping(FIELD_NAME, 2));
+        updateClusterSettings("knn.circuit_breaker.triggered", "true");
+
+        Float[] vector  = {6.0f, 6.0f};
+        ResponseException ex = expectThrows(
+                ResponseException.class, () -> addKnnDoc(INDEX_NAME, "1", FIELD_NAME, vector));
+        assertThat(EntityUtils.toString(ex.getResponse().getEntity()),
+                containsString("Cannot update KNN vector field when circuit breaker is triggered."));
+    }
```
vamshin, line 57: As an extension to this same test case, can we unset ```knn.circuit_breaker.triggered``` flag and assert new doc could be added? Same with below test case.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
chenqi0805: Sure.

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNESIT.java
chenqi0805: Sure.

chenqi0805:
@vamshin Thanks for review. Should be ready for a second look.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
jmazanec15: That makes sense.

jmazanec15 (approved):
LGTM!

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNESIT.java
vamshin: Sorry I should have mentioned this with my last comment. Can we also check there is no impact to searches when Circuit breaker triggers.  You can add a new test or change existing test

Steps

1. set CB triggered false
2. Add 4 docs
3. perform knn search (should be able to get the docs) 
4. set CB triggered true
5. Add 5th doc (should be rejected)
6. perform knn search (should be able to get the docs) 
7. se CB triggered false
8. Add 5th doc (should be accpeted)
9. perform knn search (should be able to get the docs)

chenqi0805: Np. I will add it. I think the test should pass as long as search does not go through the parser in vector mapper.

vamshin: thank you. Yes it should but this will act as a safety check for future code changes as well to ensure searches work fine even if CB triggered.

chenqi0805:
@vamshin Ready for another look. Looks like CI is not triggered. Could you take a look?

vamshin (approved):
LGTM! Thanks for the changes

PR #120, opened 2020-05-19
Updated CI.yml to remove running tests on docker

vamshin:
*Issue #, if available:*

*Description of changes:*
Integration tests already run on the ```ubuntu-latest``` machine. We do not need docker to run the integ tests and thus avoiding workflow to depend on ODFE docker image.

Added new action to that could help run sanity on ODFE docker images once we have new release.


.github/workflows/CI-docker-sanity.yml
```diff
@@ -0,0 +1,62 @@
+name: Build and Test k-NN
+on:
+  push:
+    branches:
```
jmazanec15: This action will still fail if latest ODFE docker version is unavailable. Is there a way we can check to see if the docker image is available before running the test?

vamshin: Good catch. For now updated to dummy branch. We will have to fix it to run on demand.
created issue to unblock the release https://github.com/opendistro-for-elasticsearch/k-NN/issues/121

jmazanec15: Instead of adding it now with dummy branch, I think it might be better off to leave it out of this PR.

vamshin: makes sense. Removed

jmazanec15 (approved):
LGTM

Issue #122, opened 2020-05-20
Create separate artifacts for JNI Library


jmazanec15:
Currently, the ODFE RPM/DEB installation process copies the KNN JNI library from S3 in a preinstallation script. This has 2 problems:
1. Many users' clusters do not have access to the internet, so the preinstall script leads to an installation failure
1. Uploading the binary library to an S3 bucket will not scale and will not support versioning. It also is inelegant.

In order to fix these issues, it makes sense to ship the library as an independent DEB/RPM and have the KNN plugin depend on the library artifacts. Additionally, if a user's system configuration is incompatible with the binary library artifacts, we should provide an easy way to build the artifacts from source.

PR #123, opened 2020-05-20
Build separate artifacts for library using CPack

jmazanec15:
*Issue #, if available:*
#122 

*Description of changes:*
This change uses CPack to generate separate RPM/DEB artifacts for the KNN JNI Library and adds the library as a dependency of the KNN Plugin.

The following steps will build the RPM and DEB packages in the `jni/packages` folder:
```
cd jni
cmake .
make package
```
2 packages are generated:
1. opendistro-knnlib-1.8.0.0-0.1_linux.x86_64.deb
1. opendistro-knnlib-1.8.0.0-0.1_linux.x86_64.rpm

Additionally, I changed the directory layout of JNI to follow a more standard C++ convention. The JNI directory now looks like this:
```
CMakeLists.txt
external/nmslib
src/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp
src/CMakeLists.txt
include/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.h
release/
packages/
```

Now, when running `make`, it will build the library in `jni/release` instead of `buildSrc`. `./gradlew build` will then build the library and move it to `buildSrc`.

Updated documentation as well.


build.gradle
```diff
@@ -148,16 +148,23 @@ task buildJniLib(type:Exec) {
     commandLine 'make'
 }
 
-test {
+task makeJniLib(type: Copy) {
     dependsOn buildJniLib
+    from "$buildDir/jni/release"
+    include "*"
```
vamshin, line 154: Minor: Should we specify to just copy the library? Making sure nothing else is copied

jmazanec15: Because the library extension is dependent on Operating System, I used wildcard here. Because only the library is placed in `jni/release` in the current build process, I believe this operation is okay.

vamshin (approved):
LGMT! Thanks for pulling knn library to a separate rpm so that it could be pulled from rpm repo with out having to download from other sources. This was an issue for the users who does not have internet connectivity in their production servers. Thanks for innovating here.

weicongs-amazon (approved):

Issue #124, opened 2020-05-22
Flaky test cases caused by Counter Enum


jmazanec15:
Currently, the plugin uses `KNNCounter` enum to keep track of different counts. For example, `GRAPH_QUERY_REQUESTS` keeps track of the number of queries made to graphs on a single node.

Because Unit tests run in the same JVM, this can lead to flaky failures, where the counter is incremented by some action in one test, and the value is checked in another test.

In order to fix this issue, these counters should be reset before running unit tests.

PR #125, opened 2020-05-22
Fix test structure

jmazanec15:
*Issue #, if available:*
#124 

*Description of changes:*
This PR creates a KNNTestCase from which all unit tests inherit from. This test case adds a method that resets all of the counters before execution. This will make testing independent of gradle test order.

Additionally, KNNRestTestCase and KNNResult were incorrectly placed in the `src/test/java/com/amazon/opendistroforelasticsearch/knn/index`. These two classes are not `index` specific, so I moved them to `src/test/java/com/amazon/opendistroforelasticsearch/knn`.


vamshin (approved):
LGTM!

Issue #116, opened 2020-05-19
Ability to pass ef_search parameter in the query


vamshin:
Currently ef_search is a static setting passed at index creation time. ef_search is something that could be changed per query. This issue is to provide support to mention ef_search in the query along with the ```k```.  If ef_search is not specified in the query, fall back to the index settings. 

Example query would look like

```
"query": {
  "knn": {
   "my_vector": {
     "vector": [3, 4],
     "k": 2,
     "ef_search": 1024
   }        
  }
 }
```

chenqi0805:
TODO:

1. Add `ef_search` into KNNQuery
2. Adapt loadIndex API to receive `ef_search` from query.
3. Add test case in `KNNESIT`

Issue #68, opened 2020-03-16
API to flush indices out of cache


jmazanec15:
For testing and debugging a KNN based cluster, it would be helpful to add an API that allows you to specify an index that you want removed from memory. Currently, the only way to do this is to delete the index, or setup a timebased eviction policy that will eventually remove the index.

jmazanec15:
This API should be used for RestTestCases to clear the cache before they run: https://github.com/opendistro-for-elasticsearch/k-NN/pull/183.

PR #126, opened 2020-05-22
ENH: ability to dynamically update efSearch setting

chenqi0805:
*Issue #, if available:*
#116 

*Description of changes:*
1. Add ef_search field into `KNNQueryBuilder` and `KNNQuery` and modified the relevant APIs
2. Pass in ef_search from query when getting/loading index for query
3. Modified both UnitTest and IntegrationTests to cover the enhancement 


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -273,11 +275,12 @@ public void evictAllGraphsFromCache() {
      *
      * @param indexPathUrl path for serialized hnsw graph
      * @param indexName index name
+     * @param efSearch HNSW parameter that represents "the size of the dynamic list for the nearest neighbors"
      * @return KNNIndex holding the heap pointer of the loaded graph
      * @throws Exception Exception could occur when registering the index path
      * to Resource watcher or if the JNI call throws
      */
-    public KNNIndexCacheEntry loadIndex(String indexPathUrl, String indexName) throws Exception {
+    public KNNIndexCacheEntry loadIndex(String indexPathUrl, String indexName, int efSearch) throws Exception {
```
vamshin: This method would be called only once when there is cache miss. The ```efsearch``` value passed is stored in the 
[IndexWrapper](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/jni/src/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp#L49) as part of [init](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/jni/src/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp#L204) call. So ef_search passed in subsequent queries would not be reflected. 

Ideal way is to plumb the ef_search through [queryIndex](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/v1736/KNNIndex.java#L71) function from [KNNWeight](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java#L101). As part of this function in the [jni](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/jni/src/com_amazon_opendistroforelasticsearch_knn_index_v1736_KNNIndex.cpp#L147), we could set the  ```ef_search``` using```setQueryTimeParams``` if ```ef_search``` is not null otherwise just use the existing value. 

We need to be careful while calling ```setQueryTimeParams``` because at the same time it is possible some query is already referencing this object and could cause segmentation faults. We need to kind of ensure only one thread is setting/getting ```ef_search``` .

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -290,7 +293,8 @@ public KNNIndexCacheEntry loadIndex(String indexPathUrl, String indexName) throw
         // the entry
         fileWatcher.init();
 
-        final KNNIndex knnIndex = KNNIndex.loadIndex(indexPathUrl, getQueryParams(indexName), KNNSettings.getSpaceType(indexName));
+        final KNNIndex knnIndex = KNNIndex.loadIndex(
```
vamshin: We do not have to change this method.  This method can always read the default setting so that ```IndexWrapper``` can refer it when ```ef_search``` is null.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
chenqi0805: > Ideal way is to plumb the ef_search through queryIndex function from KNNWeight. As part of this function in the jni, we could set the ef_search using setQueryTimeParams if ef_search is not null otherwise just use the existing value.

Thanks for pointing that out. I need to renew license on CLion for getting this part done.

> We need to be careful while calling setQueryTimeParams because at the same time it is possible some query is already referencing this object and could cause segmentation faults. We need to kind of ensure only one thread is setting/getting ef_search .

Thanks for reminder. My first thought is to use writeLock in queryIndex if ef_search is not null. Might need to put some thought into this. Suggestion is welcome.

jschmitz28: I suspect that supporting a dynamic efSearch parameter would either require:
1. Synchronizing search queries from multiple threads onto a shared index (which would likely hurt query performance significantly)
2. Require changes to the KNN library dependency to make the efSearch parameter a local parameter of the KNN query structure rather than a member of the index. It's currently a class member of the [HNSW index](https://github.com/nmslib/nmslib/blob/master/similarity_search/include/method/hnsw.h#L519), which is why the opendistro plugin is setting it once on index load rather than supporting per query, since mutations within queries would affect unrelated queries (see [here](https://github.com/nmslib/nmslib/blob/master/similarity_search/src/method/hnsw.cc#L1205) for one example usage of the member variable).

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -144,11 +144,12 @@ private void onRemoval(RemovalNotification<String, KNNIndexCacheEntry> removalNo
      *
      * @param key indexPath where the serialized hnsw graph is stored
      * @param indexName index name
+     * @param efSearch HNSW parameter that represents "the size of the dynamic list for the nearest neighbors"
```
jmazanec15: I would say if you include quotes include the source. Also, I think this description can be expanded, or a link to the NMSLIB documentation that discusses it would be helpful. Here is a [link](https://github.com/nmslib/nmslib/blob/master/manual/methods.md#graph-based-search-methods-sw-graph-and-hnsw)

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
vamshin: Agree. Synchronizing could hurt performance because of locking. At this point to me, Approach 2 mentioned by @jschmitz28  seems to be best way to achieve this.

Other solution i could think of is, make ```ef_search``` setting dynamic Elasticsearch index setting(currently its static index setting, once define cannot be changed). On the change of this setting, we could evict the graphs from memory and load freshly with the new ```ef_search```.  We already do this eviction when we change few settings. Example [here](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java#L203)
-  This removes the indexing cost to rebuild the whole index which is a pain point today and on next query graphs should be available for search. 
- Avoid errors possible from mutation.

chenqi0805: @vamshin I have two more points to clarify regarding your solution, first UX and second implementation:

(1) dynamical setting still needs to be updated through index settings API: https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html
so I think UX is going to be different from #116 . Is that OK? Otherwise I will return to approach 2 by @jschmitz28 .

(2) On trying to implemnt your solution by defining ef_search to be index scope and dynamic. As you mentioned, we also need to add custom setting update consumer for that ef_search. Now the issue is that IndexModule has not yet been passed into KNNSettings so we do not have way to add the consumer for index scope settings. 

From my observation, we can implement

https://github.com/elastic/elasticsearch/blob/853dd1b873d02d04231c32722d4784c9da1068dc/server/src/main/java/org/elasticsearch/plugins/Plugin.java#L131

in KNNPlugin, then we add an accept index module API in KNNSetting. Thought?

vamshin: @chenqi0805 

(1) Seems to be perfectly ok to change that. 
Regarding the approach mentioned by @jschmitz28  we could create an issue in the nmslib and ask their opinion about this change? Not sure if they had to do this way for some optimizations.

(2) Makes sense.

chenqi0805:
TODO: device test cases for dynamic ef_search.

chenqi0805:
@vamshin Do not have a good strategy to test Cache rebuilt on updating index setting. Otherwise the PR is ready for another look.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
```diff
@@ -396,4 +398,15 @@ public void setClusterService(ClusterService clusterService) {
...
         }
     }
+
+    public void onIndexModule(IndexModule module) {
+        module.addSettingsUpdateConsumer(
+                INDEX_KNN_ALGO_PARAM_EF_SEARCH_SETTING,
+                newVal -> {
+                    logger.debug("The value of setting [{}] changed to [{}]", KNN_ALGO_PARAM_EF_SEARCH, newVal);
+                    latestSettings.put(KNN_ALGO_PARAM_EF_SEARCH, newVal);
+                    // spawn a thread
```
vamshin: we could remove this comment. Cache rebuild already seem to happen in a [dedicated thread ](https://github.com/opendistro-for-elasticsearch/k-NN/blob/4f360f2e095f5ec5eaac3d4a069e03b7201e3f03/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java#L112-L116)

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
```diff
@@ -396,4 +398,15 @@ public void setClusterService(ClusterService clusterService) {
...
+                    KNNWeight.knnIndexCache.rebuild();
```
vamshin, line 409: rebuild whole cache might not be necessary as we could lose graphs of other indices. Once we have the clear cache api https://github.com/opendistro-for-elasticsearch/k-NN/issues/68 , we could selectively evict entries of a particular index. We could leave a ```TODO```  to fix after clear cache api is built. Would be great to create an issue as well to just track the fix?

vamshin (commented):
> @vamshin Do not have a good strategy to test Cache rebuilt on updating index setting. Otherwise the PR is ready for another look.

I am thinking of following scenario to test it(not the best one though).  

step1: Create index, add  1 doc, search index.
step2: read _knn/stats and note hit count say x, miss count say y
step3: change ef_search for this index (cache rebuild should happen which means entry evicted.)
step4: search index (this should be miss as cache is empty)
step5:  read _knn/stats and assert hit count = x, miss count = y+1

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
jmazanec15: Yes, I agree once we have functionality to manually remove and load an index's graphs, we should implement a `reload` function that operates at the index level. For now, a `TODO` will be fine.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
```diff
@@ -396,4 +398,15 @@ public void setClusterService(ClusterService clusterService) {
             }
         }
     }
+
+    public void onIndexModule(IndexModule module) {
+        module.addSettingsUpdateConsumer(
+                INDEX_KNN_ALGO_PARAM_EF_SEARCH_SETTING,
+                newVal -> {
+                    logger.debug("The value of setting [{}] changed to [{}]", KNN_ALGO_PARAM_EF_SEARCH, newVal);
```
jmazanec15: Nit: Can you prepend log statement with "[KNN] " so that it is easier to identify when going through logs?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
```diff
@@ -396,4 +398,15 @@ public void setClusterService(ClusterService clusterService) {
             }
         }
     }
+
+    public void onIndexModule(IndexModule module) {
+        module.addSettingsUpdateConsumer(
+                INDEX_KNN_ALGO_PARAM_EF_SEARCH_SETTING,
+                newVal -> {
```
jmazanec15, line 405: Just curious, what happens if a invalid parameter is passed in? Does it fail gracefully?

jmazanec15 (commented):
Thanks for working on this. Just a few minor comments.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
chenqi0805: Good point to check. I expect the validator in INDEX_KNN_ALGO_PARAM_EF_SEARCH_SETTING should handle it.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
chenqi0805: Thank you for comment. Good to remember that although the current line is outdated.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
chenqi0805: I have expanded testUpdateIndexSetting to cover invalid update now.

chenqi0805:
@jmazanec15 @vamshin Ready for another look.

I added `testKNNStatsAfterUpdateIndexSetting` to cover cache miss after update index setting. It is a little different from vamshin's proposal as I noticed that KNN stats(misses, hits) gets reset after updating index setting. But it still makes sure the first search after update misses.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
jmazanec15: Great thanks

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNESSettingsTestIT.java
```diff
@@ -87,5 +91,73 @@ public void testCreateIndexWithInvalidSpaceType() throws IOException {
...
+        assertEquals("400", getIndexSettingByName(INDEX_NAME, KNNSettings.KNN_ALGO_PARAM_EF_SEARCH));
+
+        Exception ex = expectThrows(ResponseException.class,
+                () -> updateIndexSettings(INDEX_NAME,
+                        Settings.builder().put(KNNSettings.KNN_ALGO_PARAM_EF_SEARCH, 1)));
+        assertThat(ex.getMessage(),
+                containsString("Failed to parse value [1] for setting [index.knn.algo_param.ef_search] must be >= 2"));
+    }
+
+    public void testKNNStatsAfterUpdateIndexSetting() throws IOException {
```
jmazanec15: I think the purpose of this test case is to confirm that the Index is evicted from the cache after the settings are updated. That being said, I think the test should be renamed `testCacheRebuiltAfterUpdateIndexSettings`.

Further more, instead of comparing deltas of hit and miss count, you can check that the cache is empty with the `indices_in_cache` stat. Then, you can just
1. Create index
2. Index sample docs
3. Search index
4. Confirm index is present in cache
5. Update index settings
6. Confirm there are no indices in the cache.

chenqi0805: Cool. Fixed.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
chenqi0805: https://github.com/nmslib/nmslib/issues/442

jmazanec15 (approved):
LGTM! Thanks for working on this @chenqi0805!

vamshin (approved):
LGTM! This feature is going to save tons of time avoiding reindexing and building graph when ever we need to experiment with ```ef_search```. Thank you.

PR #130, opened 2020-05-31
FIX: copy shared library

chenqi0805:
*Issue #, if available:*
None

*Description of changes:*
Fixes parent directory in makeJniLib gradle task. In the current master running the task leads to

```
Task :makeJniLib NO-SOURCE
```


jmazanec15 (approved):
Good catch thanks! LGTM

PR #132, opened 2020-06-02
added github action to build library artifacts

jmazanec15:
*Issue #, if available:*
#122 
*Description of changes:*
Added job in CD workflow to build DEB and RPM artifacts and copy them to S3.

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -1,5 +1,7 @@
 name: Build and Release k-NN
 on:
+  repository_dispatch:
+    types: [build-and-release-kNN]  
   push:
     tags:
       - v*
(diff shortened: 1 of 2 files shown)


.github/workflows/CD.yml
```diff
@@ -29,7 +30,24 @@ jobs:
         with:
           java-version: ${{ matrix.java }}
 
-      - name: Run build
+      - name: Build library
```
vamshin: Nit: Should we call it "Build library artifacts"?

weicongs-amazon (approved):

.github/workflows/CD.yml
jmazanec15: You are right. Because they also send artifacts to S3, I changed name to Build and ship library artifacts

vamshin (approved):
LGTM!

Issue #146, opened 2020-06-23
Upgrade Plugin to Elasticsearch 7.8.0


jmazanec15:
Upgrade plugin to run on Elasticsearch 7.8.0.

PR #147, opened 2020-06-24
Elasticsearch 7.8 support

vamshin:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/k-NN/issues/146

*Description of changes:*
Upgrade plugin to run on Elasticsearch 7.8.0.

### DIFF
--- a/build-tools/knnplugin-coverage.gradle
+++ b/build-tools/knnplugin-coverage.gradle
@@ -74,7 +74,7 @@ allprojects{
         jacocoTestReport.dependsOn integTest.runner
 
         testClusters.integTest {
-            jvmArgs " ${dummyIntegTest.jacoco.getAsJvmArg()}"
+            jvmArgs " ${dummyIntegTest.jacoco.getAsJvmArg()}".replace('javaagent:','javaagent:/')
             systemProperty 'com.sun.management.jmxremote', "true"
             systemProperty 'com.sun.management.jmxremote.authenticate', "false"
             systemProperty 'com.sun.management.jmxremote.port', "7777"
--- a/build.gradle
+++ b/build.gradle
@@ -15,7 +15,7 @@
 
 buildscript {
     ext {
-        es_version = System.getProperty("es.version", "7.7.0")
+        es_version = System.getProperty("es.version", "7.8.0")
         es_group = "org.elasticsearch"
         distribution = 'oss-zip'
     }
@@ -40,7 +40,7 @@ plugins {
     id 'java-library'
     id 'nebula.ospackage' version "8.3.0" apply false
     id 'idea'
-    id 'jacoco'
+    //id 'jacoco'
 }
 
 apply plugin: 'elasticsearch.esplugin'
@@ -52,12 +52,12 @@ def integTestDependOnJniLib = false
 if (!usingRemoteCluster) {
     integTestDependOnJniLib = true
     if (!usingMultiNode) {
-        apply from: 'build-tools/knnplugin-coverage.gradle'
+       // apply from: 'build-tools/knnplugin-coverage.gradle'
     }
 }
 
 ext {
-    opendistroVersion = '1.8.0'
+    opendistroVersion = '1.9.0'
     isSnapshot = "true" == System.getProperty("build.snapshot", "true")
     projectSubstitutions = [:]
     licenseFile = rootProject.file('LICENSE.TXT')
@@ -75,11 +75,11 @@ allprojects {
 }
 
 
-jacoco {
-    toolVersion = "0.8.5"
-}
-
-check.dependsOn jacocoTestReport
+//jacoco {
+//    toolVersion = "0.8.5"
+//}
+//
+//check.dependsOn jacocoTestReport
 
 esplugin {
     name 'opendistro-knn'
@@ -202,8 +202,7 @@ forbiddenPatterns {
 // see https://github.com/elastic/elasticsearch/blob/323f312bbc829a63056a79ebe45adced5099f6e6/buildSrc/src/main/java/org/elasticsearch/gradle/precommit/TestingConventionsTasks.java
 // enable testingConventions check will cause errors like:  "Classes ending with [Tests] must subclass [LuceneTestCase]"
 testingConventions.enabled = false
-licenseHeaders.enabled = true
-
+validateNebulaPom.enabled = false
 apply plugin: 'nebula.ospackage'
 
 // This is afterEvaluate because the bundlePlugin ZIP task is updated afterEvaluate and changes the ZIP name to match the plugin name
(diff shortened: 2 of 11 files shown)


build.gradle
```diff
@@ -40,7 +40,7 @@ plugins {
     id 'java-library'
     id 'nebula.ospackage' version "8.3.0" apply false
     id 'idea'
-    id 'jacoco'
+    //id 'jacoco'
```
jmazanec15: Why is this commented out?

build.gradle
```diff
@@ -52,12 +52,12 @@ def integTestDependOnJniLib = false
 if (!usingRemoteCluster) {
     integTestDependOnJniLib = true
     if (!usingMultiNode) {
-        apply from: 'build-tools/knnplugin-coverage.gradle'
+       // apply from: 'build-tools/knnplugin-coverage.gradle'
```
jmazanec15: Why is this commented out?

build.gradle
```diff
@@ -52,12 +52,12 @@ def integTestDependOnJniLib = false
...
     }
 }
 
 ext {
-    opendistroVersion = '1.8.0'
+    opendistroVersion = '1.9.0'
```
jmazanec15: Because this is just upgrading ES version, ODFE version does not need to be bumped up.

build.gradle
```diff
@@ -75,11 +75,11 @@ allprojects {
 }
 
 
-jacoco {
-    toolVersion = "0.8.5"
-}
-
-check.dependsOn jacocoTestReport
+//jacoco {
```
jmazanec15: Is this WIP?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -49,7 +48,7 @@
 /**
  * Field Mapper for KNN vector type.
  */
-public class KNNVectorFieldMapper extends FieldMapper implements ArrayValueMapperParser {
```
jmazanec15, line 52: Why can we remove this?

vamshin: ArrayValueMapperParser is removed in ES version 7.8. So refactored around that

build.gradle
vamshin: makes sense. Will revert it.

build.gradle
vamshin: Good catch. Forgot to fix the jacoco. Its WIP

build-tools/knnplugin-coverage.gradle
```diff
@@ -74,7 +74,7 @@ allprojects{
         jacocoTestReport.dependsOn integTest.runner
 
         testClusters.integTest {
-            jvmArgs " ${dummyIntegTest.jacoco.getAsJvmArg()}"
+            jvmArgs " ${dummyIntegTest.jacoco.getAsJvmArg()}".replace('javaagent:','javaagent:/')
```
jmazanec15, line 77: Just curious, what does this do?

build.gradle
```diff
@@ -209,8 +209,7 @@ forbiddenPatterns {
 // see https://github.com/elastic/elasticsearch/blob/323f312bbc829a63056a79ebe45adced5099f6e6/buildSrc/src/main/java/org/elasticsearch/gradle/precommit/TestingConventionsTasks.java
 // enable testingConventions check will cause errors like:  "Classes ending with [Tests] must subclass [LuceneTestCase]"
 testingConventions.enabled = false
-licenseHeaders.enabled = true
```
jmazanec15, line 212: Why was this removed?

build.gradle
```diff
@@ -231,7 +230,7 @@ afterEvaluate {
         dirMode 0755
 
         requires('elasticsearch-oss', versions.elasticsearch, EQUAL)
-        requires("opendistro-knnlib",  "1.8.0.0",  EQUAL)
+        requires("opendistro-knnlib",  "${opendistroVersion}.0",  EQUAL)
```
jmazanec15, line 226: Nice, will save us some trouble

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -235,6 +234,11 @@ public Query termQuery(Object value, QueryShardContext context) {
         }
     }
 
+    @Override
+    public final boolean parsesArrayValue() {
```
jmazanec15, line 238: Curious, what does this do?

jmazanec15 (commented):
Why is jnilib getting updated? If no change to JNI layer, that should not be changed.

build-tools/knnplugin-coverage.gradle
vamshin: Fixes jacocoagent path to work with gradle 6.5 for integration tests

build.gradle
vamshin: this was a duplicate

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
vamshin: It tells the parser to expect array. Otherwise, it treats as a single value and fails validating the vector with respect to number of dimensions

vamshin:
> Why is jnilib getting updated? If no change to JNI layer, that should not be changed.

Could be because of difference in the Mac OS, i recently updated. Anyways this library should be agnostic as we build one on the fly. We could remove this from the repo as part of different pr

build.gradle
jmazanec15: Oh cool

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
jmazanec15: Oh interesting. Thanks

jmazanec15:
> > Why is jnilib getting updated? If no change to JNI layer, that should not be changed.
> 
> Could be because of difference in the Mac OS, i recently updated. Anyways this library should be agnostic as we build one on the fly. We could remove this from the repo as part of different pr

Okay, I think the binary should not be included in this PR because this PR does not relate to JNI layer, but it is not a blocker. I will approve.

jmazanec15 (approved):

vamshin:
> > > Why is jnilib getting updated? If no change to JNI layer, that should not be changed.
> > 
> > 
> > Could be because of difference in the Mac OS, i recently updated. Anyways this library should be agnostic as we build one on the fly. We could remove this from the repo as part of different pr
> 
> Okay, I think the binary should not be included in this PR because this PR does not relate to JNI layer, but it is not a blocker. I will approve.

Rebased with master to consume PR that ignores the shared library artifacts. We should not be seeing any confusion going further.

jmazanec15 (approved):

Issue #148, opened 2020-06-25
Remove/depricate shared library in buildSrc


vamshin:
Library (libKNNIndexV1_7_3_6.jnilib, libKNNIndexV1_7_3_6.so) should be build locally to the machine running k-NN plugin to avoid compatibility issues.

vamshin:
https://github.com/opendistro-for-elasticsearch/k-NN/pull/149

PR #149, opened 2020-06-25
deprecate/remove shared library from buildSrc

vamshin:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/k-NN/issues/148

*Description of changes:*

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ build
 *.log
 out/
 buildSrc/es-knn-offline-repo-*
+buildSrc/libKNNIndexV1_7_3_6*
 oss/*
 *.iml
 jni/CMakeCache.txt
--- a/build.gradle
+++ b/build.gradle
@@ -124,28 +124,21 @@ task buildJniLib(type:Exec) {
     commandLine 'make'
 }
 
-task makeJniLib(type: Copy) {
-    dependsOn buildJniLib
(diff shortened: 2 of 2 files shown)


jmazanec15 (commented):
Thanks for making this PR. Just curious, do you think [this task](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/build.gradle#L127) is needed for moving the library from `jni/build/release` to `buildSrc`, or could we just change this [env variable](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/build.gradle#L137) in all of the places that set it to `jni/build/release`

I guess my question is why should library be in `buildSrc` at all?

vamshin:
> Thanks for making this PR. Just curious, do you think [this task](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/build.gradle#L127) is needed for moving the library from `jni/build/release` to `buildSrc`, or could we just change this [env variable](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/build.gradle#L137) in all of the places that set it to `jni/build/release`
> 
> I guess my question is why should library be in `buildSrc` at all?

No specific reason actually. Makes sense to remove the ```task``` and rely on ```jni/release``` to look out for library

jmazanec15 (approved):
LGTM thanks

Issue #154, opened 2020-07-01
Bad recall from ODFE1.8 


vamshin:
Problem:- 
Bad recall with repeated scores for vectors with different magnitude.

Issue:-
Elasticsearch upgraded Lucene version to 8.5.1 from 7.7 version. This release of Lucene introduced change for [compressed Binary doc values](https://github.com/apache/lucene-solr/commit/ce2959fe4cb1d1e77df04464c46004bf7846f6b5). As part of this change,  BinaryDecoder is added which returns the whole byte array for binary doc values with exact offset (position where the vector is stored) and length of bytes to be read, instead of slicing the byte array with just the bytes needed to decode vector.  

Before Lucene 8.5 (Elasticsearch version < 7.7)
https://github.com/apache/lucene-solr/blob/releases/lucene-solr/8.4.1/lucene/core/src/java/org/apache/lucene/codecs/lucene80/Lucene80DocValuesProducer.java#L688

From Lucene 8.5 (Elasticsearch version >= 7.7)
https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/codecs/lucene80/Lucene80DocValuesProducer.java#L838-L840

Our [logic](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNNCodecUtil.java#L48) to extract the float values(vector) from the ByteRef depends on sliced byte array and hence offset always points to 0 and length which is the size of sliced array.

Because of this issue, we always read the first vector for the whole compressed block and pass on the same vector to hnsw index resulting in duplicate points and scores which resulting in bad recall. This will be evident after the forcemerge

This fix needs to be backported to odfe1.8 and odfe1.9

vamshin:
Fix here https://github.com/opendistro-for-elasticsearch/k-NN/pull/155/

tanguilp:
Hi Vashmi! I've setup ODFE1.8 from the Docker image, and loaded a few thousand pieces of data with associated vectors.

When performing search requests, the same vector is always returned, and I get only one result.

Couldn't it be somehow related to this bug?

vamshin:
@tanguilp you are right. This is related to this bug. We are in the process of updating artifacts

vamshin:
This issue is now fixed in ODFE1.9

timforr:
Was this fixed in ODFE1.8?

[...]

PR #155, opened 2020-07-01
Recall bug fix for odfe>=1.8

vamshin:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/k-NN/issues/154

*Description of changes:*
Details in the issue link above


jmazanec15 (approved):
LGTM! Great work finding this fix.

As a note, to confirm that it works, I ran the following test:
1. Ingested the SIFT data set from [ANN Benchmarks](https://github.com/erikbern/ann-benchmarks/#data-sets) into a single node instance that used the 1.8.0.0 Community AMI with 5 shards and default algorithm settings
2. Force merged the index to one segment
3. Ran the 10,000 test queries with size and k = 100, and computed recall. Before the fix,  recall < 0.75. After the fix, recall > 0.99

Issue #158, opened 2020-07-06
Flaky unit tests due to cache state


jmazanec15:
Related to [this action failure](https://github.com/opendistro-for-elasticsearch/k-NN/runs/785598865?check_suite_focus=true)

Cache is not reset before running unit tests. For example, in [KNNIndexCacheTests](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCacheTests.java#L46). This can lead to leftover state across tests, causing failures. This needs to be cleaned up.

PR #159, opened 2020-07-06
Reset state for uTs so tests run independently

jmazanec15:
*Issue #, if available:*
#158 

*Description of changes:*
Test cases were failing because the state from the KNNIndexCache (singleton) was spilling over from other test cases. This PR prevents this by adding a tearDown method which resets counter, clears out the cache, and closes the cache. 

In order to test, I replicated the failure on an Ubuntu 16 machine. Then, with the change, it passes. Additionally, I tested the change on a Mac. I was not able to reproduce the failure, but, with the fix, `./gradlew build` still passes.


vamshin (commented):
LGTM!

Issue #135, opened 2020-06-12
Introduce API to Load Graphs into Cache


jmazanec15:
Initial queries can experience high latencies because graphs need to be loaded into cache before they can be searched. As a work around, users can run queries for a warmup period in order to load graphs into memory. However, this approach can be inefficient, tedious, and error prone.

We should introduce an API that performs this functionality instead. The API should have the following interface:
`GET /_opendistro/_knn/warmup/<index 1>,<index 2>,...`

When called, it will load the graphs into memory for search.

PR #162, opened 2020-07-09
Add Warmup API to load indices graphs into memory

jmazanec15:
*Issue #, if available:*
#135 

*Description of changes:*
This PR adds a warmup API so that users can explicitly load graphs for their indices into memory.

The API looks like this:
```
GET /_opendistro/_knn/warmup/<index 1>,<index 2>,
{
  "_shards" : {
    "total" : 6,
    "successful" : 6,
    "failed" : 0
  }
}
```

This API uses `TransportBroadcastByNodeAction` to broadcast the request to all shards for the specified indices. In order to load a shard's graphs, we wrap `IndexShard` in `KNNIndexShard` to look up the HNSW files for that shard. Then, we take those HNSW files and load them into the cache.

Added integration and unit tests for the added components. Additionally, I created a ODFE 1.9 cluster with RPM and installed this plugin and ran tests against a larger data set.


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java
```diff
@@ -0,0 +1,124 @@
...
+ * KNNIndexShard wraps IndexShard and adds methods to perform k-NN related operations against the shard
+ */
+public class KNNIndexShard {
+    private IndexShard indexShard;
+    private KNNIndexCache knnIndexCache;
+
+    private static Logger logger = LogManager.getLogger(KNNIndexShard.class);
+
+    /**
+     * Constructor to generate KNNIndexShard. We do not perform validation that the Index the IndexShard is from
```
kaituo: remove "the Index" and just leave "the IndexShard"?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java
```diff
@@ -0,0 +1,124 @@
...
+     * @return List of HNSW Paths
+     * @throws IOException Thrown when the SegmentReader is attempting to read the segments files
+     */
+    public List<String> getHNSWPaths(IndexReader indexReader) throws IOException {
+        List<String> hnswFiles = new ArrayList<>();
+        for (LeafReaderContext leafReaderContext : indexReader.leaves()) {
+            SegmentReader reader = (SegmentReader) FilterLeafReader.unwrap(leafReaderContext.reader());
+            hnswFiles.addAll(reader.getSegmentInfo().files().stream()
+                    .filter(fileName -> fileName.endsWith(getHNSWFileExtension(reader.getSegmentInfo().info)))
+                    .map(fileName -> shardPath().resolveIndex().resolve(fileName).toString())
```
kaituo, line 108: What if resolve(fileName) returns null?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/transport/KNNWarmupResponse.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.elasticsearch.action.support.DefaultShardOperationFailedException;
+import org.elasticsearch.action.support.broadcast.BroadcastResponse;
+import org.elasticsearch.common.io.stream.StreamInput;
+import org.elasticsearch.common.xcontent.ToXContentObject;
+
+import java.io.IOException;
+import java.util.List;
+
+public class KNNWarmupResponse extends BroadcastResponse implements ToXContentObject {
```
vamshin, line 30: class doc missing

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/transport/KNNWarmupRequest.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+
+package com.amazon.opendistroforelasticsearch.knn.plugin.transport;
+
+import org.elasticsearch.action.support.broadcast.BroadcastRequest;
+import org.elasticsearch.common.io.stream.StreamInput;
+
+import java.io.IOException;
+
+public class KNNWarmupRequest extends BroadcastRequest<KNNWarmupRequest> {
```
vamshin, line 26: class doc missing. You might want to take care in other places as wel

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/rest/RestKNNWarmupHandler.java
```diff
@@ -0,0 +1,88 @@
...
+    public List<Route> routes() {
+        return ImmutableList.of(
+                new Route(RestRequest.Method.GET, KNNPlugin.KNN_BASE_URI + "/warmup/{index}")
+        );
+    }
+
+    @Override
+    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) {
+        KNNWarmupRequest knnWarmupRequest = createKNNWarmupRequest(request);
+        return channel -> client.execute(KNNWarmupAction.INSTANCE, knnWarmupRequest, new RestToXContentListener<>(channel));
```
vamshin, line 78: +1 for making it async.

we could log here saying that ```warm up``` initiated. Should we capture the number of warmup requests in the knn stats?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/rest/RestKNNWarmupHandler.java
```diff
@@ -0,0 +1,88 @@
...
+
+    private KNNWarmupRequest createKNNWarmupRequest(RestRequest request) {
+        String[] indexNames = Strings.splitStringByCommaToArray(request.param("index"));
+        Index[] indices =  indexNameExpressionResolver.concreteIndices(clusterService.state(), strictExpandOpen(),
+                indexNames);
+
+        Arrays.stream(indices).forEach(index -> {
+            if (!"true".equals(clusterService.state().metadata().getIndexSafe(index).getSettings().get(KNN_INDEX))) {
+                throw new KNNInvalidIndexException(index.getName(),
+                        "Unable to create warmup index that has 'knn' setting set to false");
```
vamshin: 1.  We could update log to tell the exact setting name. How about below message

throw new KNNInvalidIndexException(index.getName(),
                        "Warmup request rejected as one of the index has '**index.knn**' setting set to false");



2. Should we loop through the indices and mention full list of indices that are invalid instead of exiting on the first error?

3. How about making ```indexNames``` from String array to ```set``` and pass this as a ```ImmutableSet``` to ```KNNWarmupRequest```. This was we could avoid duplicates.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/rest/RestKNNWarmupHandler.java
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    @Override
+    protected RestChannelConsumer prepareRequest(RestRequest request, NodeClient client) {
+        KNNWarmupRequest knnWarmupRequest = createKNNWarmupRequest(request);
+        return channel -> client.execute(KNNWarmupAction.INSTANCE, knnWarmupRequest, new RestToXContentListener<>(channel));
+    }
+
+    private KNNWarmupRequest createKNNWarmupRequest(RestRequest request) {
+        String[] indexNames = Strings.splitStringByCommaToArray(request.param("index"));
```
vamshin, line 82: if indexNames empty, return from here with proper log message?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/KNNPlugin.java
```diff
@@ -143,8 +148,10 @@
                                              Supplier<DiscoveryNodes> nodesInCluster) {
 
         RestKNNStatsHandler restKNNStatsHandler = new RestKNNStatsHandler(settings, restController, knnStats);
+        RestKNNWarmupHandler restKNNWarmupHandler = new RestKNNWarmupHandler(settings, restController, clusterService,
```
vamshin, line 151: is ```getRestHandlers``` always called after ```getMappers```? Making sure clusterService is not null.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java
```diff
@@ -0,0 +1,124 @@
...
+     * @return a List of KNNIndex's from this shard that are in the cache after this operation.
+     * @throws IOException Thrown when getting the HNSW Paths to be loaded in
+     */
+    public List<KNNIndex> warmup() throws IOException {
+        logger.info("[KNN] Warming up index: " + getIndexName());
+        Engine.Searcher searcher = indexShard.acquireSearcher("knn-warmup");
+        List<KNNIndex> indices;
+        try {
+            indices = knnIndexCache.getIndices(getHNSWPaths(searcher.getIndexReader()), getIndexName());
+            searcher.close();
```
vamshin, line 90: move ```searcher.close()``` to finally block and you could remove from ```catch```

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java
vamshin: As part of the unit test for ```getHNSWPaths,  we could add a test to create a document in the knn index with out knn vector and refresh it and check the behavior. in this case the segment does not have hnsw files and we should assert it returns empty.

vamshin (commented):
Took a quick glance. Will do one more iteration after addressing the comments.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java
jmazanec15: I think I wiil change to "We do not perform validation that the index the shard is from"

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java
jmazanec15: Good suggestion. Will update

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java
jmazanec15: resolve(fileName) should not return null because we get the fileNames from the indexReader. However, I can add a test case to confirm the behavior.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/KNNPlugin.java
jmazanec15: Do you mean createComponents, not getMappers?

And yes
1. [createComponents](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/node/Node.java#L506)
2. [getRestHandlers](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/node/Node.java#L670)

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/rest/RestKNNWarmupHandler.java
jmazanec15: Will add this check.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/rest/RestKNNWarmupHandler.java
jmazanec15: 1. Makes sense
2. Yes, I think that would be better design.
3. Yes that makes sense. Will change that.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/transport/KNNWarmupRequest.java
jmazanec15: Will take care of it thanks.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/transport/KNNWarmupResponse.java
jmazanec15: Will take care.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/KNNPlugin.java
vamshin: Sorry. I meant ```CreateComponent```. Thanks for confirming.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/rest/RestKNNWarmupHandler.java
jmazanec15: I can add the log statement. I am not sure if it is useful to know how many warmup requests there have been.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java
jmazanec15: Actually, fileName has the `@NotNull` annotation. So it cannot be null.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/rest/RestKNNWarmupHandler.java
jmazanec15: For Point 3, `KNNWarmupRequest` extends `BroadcastRequest` which takes an array of indices. So, I do not think I should modify that. Additionally, `KNNWarmupTransportAction` extends `TransportBroadcastByNodeAction`, which deduplicates the indices [here](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/action/support/broadcast/node/TransportBroadcastByNodeAction.java#L252), so I am not sure change 3 is necessary.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/rest/RestKNNWarmupHandler.java
jmazanec15: Actually, because I only register [this route](https://github.com/opendistro-for-elasticsearch/k-NN/pull/162/files#diff-9b2b661b10b3b79228d72391a36fd8edR64), it is impossible to pass in no indices. Instead, it fails with the following:

```
curl ${HOST_NAME}/_opendistro/_knn/warmup?pretty
{
  "error" : {
    "root_cause" : [
      {
        "type" : "illegal_argument_exception",
        "reason" : "request [/_opendistro/_knn/warmup] contains unrecognized parameter: [nodeId]"
      }
    ],
    "type" : "illegal_argument_exception",
    "reason" : "request [/_opendistro/_knn/warmup] contains unrecognized parameter: [nodeId]"
  },
  "status" : 400
}
```

It actually does not reach this rest handler. Is this okay?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java
kaituo: Could you point me where annotation is?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/rest/RestKNNWarmupHandler.java
kaituo: For 1, do we want to return the index name that has 'index.knn' setting to be false?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java
jmazanec15: I see it in the source code in intellij. I think Intellij may have added the annotation.

image

I am not sure what this implies. Do you know what this does @kaituo ?

kaituo: maybe https://www.jetbrains.com/help/idea/nullable-and-notnull-annotations.html#notnull

kaituo (approved):

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/rest/RestKNNWarmupHandler.java
vamshin: makes sense. Thanks

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/rest/RestKNNWarmupHandler.java
vamshin: Correct. Actually list of index names that have ```index.knn``` setting to false.

README.md
```diff
@@ -404,6 +404,35 @@ GET /_opendistro/_knn/HYMrXXsBSamUkcAjhjeN0w/stats/circuit_breaker_triggered,gra
 }
 ```
 
+## Warmup API
+### Overview
+The HNSW graphs used to perform k-Approximate Nearest Neighbor Search are stored as `.hnsw` files with the other Lucene segment files. In order to perform search on these graphs, they need to be loaded into memory. If the graphs have not yet been loaded into memory, upon search, they will first be loaded and then searched. This can cause high latency during initial queries. To avoid this, users will often run random queries during a warmup period. After this warmup period, the graphs will be loaded into memory and their production workloads can begin. This process is indirect and requires extra effort.
```
vamshin: Minor: In places where we mention ```memory``` should we be explicit calling ```native memory``` or ```off-heap``` to set context where the graphs are actually loaded?

jmazanec15: Sure, will update

README.md
```diff
@@ -404,6 +404,35 @@ GET /_opendistro/_knn/HYMrXXsBSamUkcAjhjeN0w/stats/circuit_breaker_triggered,gra
 }
 ```
 
+## Warmup API
+### Overview
+The HNSW graphs used to perform k-Approximate Nearest Neighbor Search are stored as `.hnsw` files with the other Lucene segment files. In order to perform search on these graphs, they need to be loaded into native memory. If the graphs have not yet been loaded into native memory, upon search, they will first be loaded and then searched. This can cause high latency during initial queries. To avoid this, users will often run random queries during a warmup period. After this warmup period, the graphs will be loaded into native memory and their production workloads can begin. This process is indirect and requires extra effort.
```
vamshin, line 409: Should we also call out Idempotency with the ```warmup``` api? warmup api ensures only the graphs belonging to the segments that did not get loaded are loaded to memory. Existing graphs would be untouched?

jmazanec15: Sure, I can call this out.

vamshin (approved):
LGTM! This api is super useful to limit latencies on the initial queries.  Thanks for the changes.

Issue #163, opened 2020-07-13
Opendistro 1.9 crashes for certain instance types RPM/DEB/TAR


jmazanec15:
# Overview
There is a bug for k-NN binary library for ODFE 1.9 for RPM/TAR/DEB artifacts. The library crashes ES on specific machines. For instance, ODFE 1.9 does not crash on m5 and c5 AWS instance types work, however, ODFE 1.9 crashes on m4 instance types.

This is may be due to the environment in which the binary library we distribute is compiled. Currently, we build the library using Github Actions runners. It is possible the binary is compiled to use optimized instructions available on the machine the Github runner uses, which is not available on certain instance types. The binary that we distribute should not contain optimized instructions in order to maximize compatibility.

# Repro Steps
1. Launch AWS EC2 instance with type m4.xlarge
2. Following this link to install and start RPM or TAR or DEB: https://opendistro.github.io/for-elasticsearch-docs/docs/install/rpm/
3. Run the following commands:
```
curl -X PUT "https://localhost:9200/myindex" -u admin:admin --insecure -H 'Content-Type: application/json' -d'
{
  "settings" : {
    "number_of_shards" :   1,
    "number_of_replicas" : 0,
    "index": {
        "knn": true,
    }
  },
  "mappings": {
      "properties": {
        "my_vector": {
          "type": "knn_vector",
          "dimension": 2
        }
      }
  }
}
'
 
curl -X POST "https://localhost:9200/_bulk" -u admin:admin --insecure -H 'Content-Type: application/json' -d'
{ "create" : { "_index" : "myindex", "_id": "6" } } } }
{ "my_vector" : [1.5, 2.5]}
{ "create" : { "_index" : "myindex", "_id": "3" } } } }
{ "my_vector" : [1.5, 2.5]}
{ "create" : { "_index" : "myindex", "_id": "4" } } } }
{ "my_vector" : [1.5, 2.5]}
{ "create" : { "_index" : "myindex", "_id": "5" } } } }
{ "my_vector" : [1.5, 2.5]}
{ "create" : { "_index" : "myindex", "_id": "6" } } } }
{ "my_vector" : [1.5, 2.5]}
{ "create" : { "_index" : "myindex", "_id": "7" } } } }
{ "my_vector" : [1.5, 2.5]}
{ "create" : { "_index" : "myindex", "_id": "8" } } } }
{ "my_vector" : [1.5, 2.5]}
[...]

PR #164, opened 2020-07-14
Pass -march=x86-64 to build JNI library

jmazanec15:
*Issue #, if available:*
#163 

*Description of changes:*
This PR changes the way in which we build the JNI library artifacts we release with Opendistro. Previously, we compiled the NMSLIB portion of the library with `-march=native`. The problem with this is that the machine the Github runner used, had optimized instruction sets that are not present on other machines a user might run the plugin on. This would then lead to an unexpected crash.

This PR changes `-march=native` to the more generic `-march=x86-64` compile option, providing more general compatibility for our library. However, this may lead some users to get worse performance than they otherwise expected, because the distributed binary will no longer include optimized instructions. For users to avoid this, they should build the library from source in an environment that mirrors their production environment. This PR adds this suggestion in the README as well.

In order to test, I built the RPM on an Amazon Linux 2 machine on an m5.xlarge EC2 instance and then installed it on an m4.xlarge instance. I was able to confirm that the steps from #163 that led to the crash earlier, did not lead to a crash with this method.

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -35,6 +35,8 @@ jobs:
           CXX: ${{ matrix.compiler }}
         run: |
             cd jni
+            git submodule update --init -- jni/external/nmslib
+            sed -i 's/-march=native/-march=x86-64/g' external/nmslib/similarity_search/CMakeLists.txt
             cmake .
             make package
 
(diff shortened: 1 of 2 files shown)


vamshin (approved):
LGTM!

Issue #165, opened 2020-07-16
RPM/DEB versioning wrong for knnlib artifact


jmazanec15:
The RPM/DEB versioning currently follow this format:
`opendistro-knnlib vX.X.X.X-0.X`

when they should follow this format:
`opendistro-knnlib vX.X.X.X-X`

This can be fixed [here](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/jni/CMakeLists.txt#L79) by changing it to 1.

PR #166, opened 2020-07-16
Fix versioning for lib artifacts

jmazanec15:
*Issue #, if available:*
#165 

*Description of changes:*
This change fixes the versioning for the library artifacts. Before, they had the versioning 1.9.0.0-0.1, now they have 1.9.0.0-1 versioning. This is consistent with the ODFE plugin versioning.


vamshin (approved):
LGTM!

PR #177, opened 2020-07-24
Performance tuning/Recommendations

vamshin:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/k-NN/issues/64

*Description of changes:*
Adds Performance tuning/Recommendations to improve indexing/search performance.

### DIFF
--- a/PerformanceTuning.md
+++ b/PerformanceTuning.md
@@ -8,8 +8,6 @@ In this document we provide recommendations for performance tuning to improve in
 * Each graph in the segment returns *<=k* neighbors. 
 * Coordinator node picks up final *size* number of neighbors from the neighbors returned by each shard
 
-To improve performance it is necessary to keep the number of segments under control. Ideally having 1 segment per shard will give the optimal performance with respect to search latency. We can achieve more parallelism by having more shards per index. We can control the number of segments either during indexing by asking Elasticsearch to slow down the segment creation by disabling the refresh interval or choosing larger refresh interval, increasing the flush threshold OR force-merging to 1 segment after all the indexing finishes and before searches.
-
 ##Indexing Performance Tuning
 
 The following steps could help improve indexing performance especially when you plan to index large number of vectors at once. 
@@ -30,7 +28,9 @@ The following steps could help improve indexing performance especially when you 
  ```
     Having replication set to 0, will avoid duplicate construction of graphs in 
     both primary and replicas. When we enable replicas after the indexing, the 
-    serialized graphs are directly copied. 
+    serialized graphs are directly copied. Having no replicas means that losing 
+    a node(s) may incur data loss, so it is important that the data lives elsewhere 
+    so that this initial load can be retried in case of an issue.
  ```
 More details [here](https://www.elastic.co/guide/en/elasticsearch/reference/master/tune-for-indexing-speed.html#_disable_replicas_for_initial_loads)
     
@@ -81,6 +81,11 @@ Please refer following doc (https://www.elastic.co/guide/en/elasticsearch/refere
 
 ##Search Performance Tuning
 
+To improve Search performance it is necessary to keep the number of segments under control. Lucene's IndexSearcher will search over all of the segments in a shard to find the 'size' best results. But, because the complexity of search for the HNSW algorithm is logarithmic with respect to the number of vectors, searching over 5 graphs with a 100 vectors each and then taking the top size results from ```5*k``` results will take longer than searching over 1 graph with 500 vectors and then taking the top size results from k results. 
+Ideally having 1 segment per shard will give the optimal performance with respect to search latency. We could configure index to have multiple shards to aviod having giant shards and achieve more parallelism.
+
+We can control the number of segments either during indexing by asking Elasticsearch to slow down the segment creation by disabling the refresh interval or choosing larger refresh interval.
+
 ### Warm up
 
 The graphs are constructed during indexing, but they are loaded into memory during the first search. The way search works in Lucene is that each segment is searched sequentially (so, for k-NN, each segment returns up to k nearest neighbors of the query point) and the results are aggregated together and ranked based on the score of each result (higher score --> better result). 
@@ -122,9 +127,12 @@ please refer this [page.](https://discuss.elastic.co/t/what-does-it-mean-to-stor
 ```
 ##Improving Recall 
 
-Recall could depend on multiple factors like number of dimensions, segments(searching over large number of small segments and aggregating the results leads better recall than searching over small number of large segments and aggregating results. The larger the graph the more chances of losing recall if sticking to smaller algorithm parameters. Choosing larger values for algo params should help solve this issue), number of vectors, etc.  Recall can be configured by adjusting the algorithm parameters of hnsw algorithm exposed through index settings. Algorithm params that control the recall are *m, ef_construction, ef_search*. For more details on influence of algorithm parameters on the indexing, search recall, please refer this  doc (https://github.com/nmslib/hnswlib/blob/master/ALGO_PARAMS.md).  Increasing these values could help recall(better search results) but at the cost of higher memory utilization and increased indexing time. Our default values work on a broader set of use cases from our experiments but we encourage users to run their own experiments on their data sets and choose the appropriate values. You could refer to these settings in this section (https://github.com/opendistro-for-elasticsearch/k-NN#index-level-settings). We will add details on our experiments shortly here.
+Recall could depend on multiple factors like number of vectors, number of dimensions, segments etc. Searching over large number of small segments and aggregating the results leads better recall than searching over small number of large segments and aggregating results. The larger the graph the more chances of losing recall if sticking to smaller algorithm parameters. 
+Choosing larger values for algorithm params should help solve this issue but at the cost of search latency and indexing time. That being said, it is important to understand your system's requirements for latency and accuracy, and then to choose the number of segments you want your index to have based on experimentation.
 
-Memory Estimation
+Recall can be configured by adjusting the algorithm parameters of hnsw algorithm exposed through index settings. Algorithm params that control the recall are *m, ef_construction, ef_search*. For more details on influence of algorithm parameters on the indexing, search recall, please refer this  doc (https://github.com/nmslib/hnswlib/blob/master/ALGO_PARAMS.md).  Increasing these values could help recall(better search results) but at the cost of higher memory utilization and increased indexing time. Our default values work on a broader set of use cases from our experiments but we encourage users to run their own experiments on their data sets and choose the appropriate values. You could refer to these settings in this section (https://github.com/opendistro-for-elasticsearch/k-NN#index-level-settings). We will add details on our experiments shortly here.
+
+##Memory Estimation
 
 AWS Elasticsearch Service clusters allocate 50% of available RAM in the Instance capped around 32GB (because of JVM GC performance limit). Graphs part of k-NN are loaded outside the Elasticsearch process JVM. We have circuit breakers to limit graph usage to 50% of the left over RAM space for the graphs.
 
--- a/PerformanceTuning.md
+++ b/PerformanceTuning.md
@@ -77,6 +77,7 @@ Please refer following doc (https://www.elastic.co/guide/en/elasticsearch/refere
 
 ## Search Performance Tuning
 
+### Fewer Segments
 To improve Search performance it is necessary to keep the number of segments under control. Lucene's IndexSearcher will search over all of the segments in a shard to find the 'size' best results. But, because the complexity of search for the HNSW algorithm is logarithmic with respect to the number of vectors, searching over 5 graphs with a 100 vectors each and then taking the top size results from ```5*k``` results will take longer than searching over 1 graph with 500 vectors and then taking the top size results from k results. 
 Ideally having 1 segment per shard will give the optimal performance with respect to search latency. We could configure index to have multiple shards to aviod having giant shards and achieve more parallelism.
 
--- a/PerformanceTuning.md
+++ b/PerformanceTuning.md
@@ -1,0 +1,147 @@
+#KNN Performance Tuning
+
+
+In this section we provide recommendations for performance tuning to improve indexing/search performance with the k-NN plugin.  On a high level k-NN works on following principles
+
+* Graphs are created per (Lucene) segment
+* Queries execute on segments sequentially inside the shard (same as any other Elasticsearch query). 
+* Each graph in the segment returns ‘k’ neighbors and the size results with the highest score is returned by the coordinating node. Note that size can be greater or smaller than k.
+
+To improve performance it is necessary to keep the number of segments under control. Ideally having 1 segment per shard will give the optimal performance with respect to search latency. We can achieve more parallelism by having more shards per index. We can control the number of segments either during indexing by asking Elasticsearch to slow down the segment creation by disabling the refresh interval or choosing larger refresh interval, increasing the flush threshold OR force-merging to 1 segment after all the indexing finishes and before searches.
+
+##Indexing Performance Tuning
+
+Following steps could help improve indexing performance especially when you plan to index large number of vectors at once. 
+
+* Disable refresh interval  (Default = 1 sec)
+ 
+ Disable refresh interval or set a long duration for refresh interval to avoid creating multiple smaller segments
+  ```  
+    PUT /<index_name>/_settings
+        {
+            "index" : {
+                "refresh_interval" : "-1"
+            }
+        }
+  ```
+* Disable flush
+ ```
+    Increase the "index.translog.flush_threshold_size" to some bigger value lets say "10gb", default is 512MB
+ ```
+* No Replicas (No Elasticsearch replica shard)
+ ```
+    Having replication set to 0, will avoid duplicate construction of graphs in 
+    both primary and replicas. When we enable replicas after the indexing, the 
+    serialized graphs are directly copied.
+ ```
+    
+* Increase number of indexing threads
+  ```
+    If the hardware we choose have multiple cores, we could allow multiple threads 
+    in graph construction and there by speed up the indexing process. You could determine
+    the number of threads to be alloted by using the following setting   
+    https://github.com/opendistro-for-elasticsearch/k-NN#knnalgo_paramindex_thread_qty.
+     
+    Please keep an eye on CPU utilization and choose right number of threads. Since graph
+    construction is costly, having multiple threads can put additional load on CPU. 
+  ```
+    
+* Index all docs (Perform bulk indexing)
+
+* Forcemerge 
+  
+ Forcemerge is a costly operation and could take a while depending on number of segments and size of the segments.
+ To ensure force merge is completed, we could keep calling forcemerge with 5 minute interval till you get 200 response.
+    
+    curl -X POST "localhost:9200/myindex/_forcemerge?max_num_segments=1&pretty"
+    
+* Call refresh 
+
+ Might not needed but to ensure the buffer is cleared and all segments are up. 
+ ```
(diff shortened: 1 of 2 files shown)


PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
+#KNN Performance Tuning
+
+
+In this section we provide recommendations for performance tuning to improve indexing/search performance with the k-NN plugin.  On a high level k-NN works on following principles
```
jmazanec15: replace "In this section" with "In this document,"

Also, replace "On a high level" with "From a high level," 

Also, add a ":" after principles.

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
...
+
+* Graphs are created per (Lucene) segment
+* Queries execute on segments sequentially inside the shard (same as any other Elasticsearch query).
```
jmazanec15: Remove the period at the end if it is a list item

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
+#KNN Performance Tuning
+
+
+In this section we provide recommendations for performance tuning to improve indexing/search performance with the k-NN plugin.  On a high level k-NN works on following principles
+
+* Graphs are created per (Lucene) segment
+* Queries execute on segments sequentially inside the shard (same as any other Elasticsearch query). 
+* Each graph in the segment returns ‘k’ neighbors and the size results with the highest score is returned by the coordinating node. Note that size can be greater or smaller than k.
```
jmazanec15: Same as above

Also instead of adding single quotes around k, just Italicize it like \*k\*. Do the same with size.

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
...
+
+To improve performance it is necessary to keep the number of segments under control. Ideally having 1 segment per shard will give the optimal performance with respect to search latency. We can achieve more parallelism by having more shards per index. We can control the number of segments either during indexing by asking Elasticsearch to slow down the segment creation by disabling the refresh interval or choosing larger refresh interval, increasing the flush threshold OR force-merging to 1 segment after all the indexing finishes and before searches.
+
+##Indexing Performance Tuning
+
+Following steps could help improve indexing performance especially when you plan to index large number of vectors at once.
```
jmazanec15: Start sentence with "The".

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
...
+ Disable refresh interval or set a long duration for refresh interval to avoid creating multiple smaller segments
+  ```  
+    PUT /<index_name>/_settings
+        {
+            "index" : {
+                "refresh_interval" : "-1"
+            }
+        }
+  ```
+* Disable flush
```
jmazanec15: Could you add a sentence description here?

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
...
+ ```
+    Increase the "index.translog.flush_threshold_size" to some bigger value lets say "10gb", default is 512MB
+ ```
+* No Replicas (No Elasticsearch replica shard)
```
jmazanec15: Change this to "Enable Replicas after indexing"

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
...
+ ```
+    Having replication set to 0, will avoid duplicate construction of graphs in 
+    both primary and replicas. When we enable replicas after the indexing, the
```
jmazanec15, line 28: Is there any risk associated with not having replicas during indexing?

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
...
+        }
+ ```
+
+Please refer following doc (https://www.elastic.co/guide/en/elasticsearch/reference/master/tune-for-indexing-speed.html) for more details on improving indexing performance in general.
+
+##Search Performance Tuning
+
+### Warm up
+
+The graphs are constructed during indexing, but they are loaded into memory during the first search. The way search works in Lucene is that each segment is searched sequentially (so, for k-NN, each segment returns up to k nearest neighbors of the query point) and the results are aggregated together and ranked based on the score of each result (higher score --> better result).
```
jmazanec15: I think I gave wrong information on this. I believe Lucene's `IndexSearcher` uses a threadpool to search over different segments concurrently, right?

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
+#KNN Performance Tuning
+
+
+In this section we provide recommendations for performance tuning to improve indexing/search performance with the k-NN plugin.  On a high level k-NN works on following principles
+
+* Graphs are created per (Lucene) segment
+* Queries execute on segments sequentially inside the shard (same as any other Elasticsearch query). 
+* Each graph in the segment returns ‘k’ neighbors and the size results with the highest score is returned by the coordinating node. Note that size can be greater or smaller than k.
+
+To improve performance it is necessary to keep the number of segments under control. Ideally having 1 segment per shard will give the optimal performance with respect to search latency. We can achieve more parallelism by having more shards per index. We can control the number of segments either during indexing by asking Elasticsearch to slow down the segment creation by disabling the refresh interval or choosing larger refresh interval, increasing the flush threshold OR force-merging to 1 segment after all the indexing finishes and before searches.
```
jmazanec15: So, say we have a somewhat sizeable shard, but within the Elasticsearch recommendation. Say 20 GB. In this case, it probably wouldn't make sense to have a single segment right? Would it be better to discuss the tradeoff instead of just saying 1 segment? 

"Lucene's IndexSearcher will search over all of the segments in a shard to find the `size` best results. Searching over a large amount of smaller graphs and then selecting the top `size` results may result in higher recall than just searching over a smaller number of larger graphs. But, because the complexity of search for the HNSW algorithm is logarithmic with respect to the number of vectors, searching over 5 graphs with a 100 vectors each and then taking the top `size` results from `5*k` results will take longer than searching over 1 graph with 500 vectors and then taking the top `size` results from `k` results (ignoring concurrency for the sake of example). That being said, it is important to understand your system's requirements for latency and accuracy, and then to choose the number of segments you want your index to have based on experimentation.

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
...
+* No Replicas (No Elasticsearch replica shard)
+ ```
+    Having replication set to 0, will avoid duplicate construction of graphs in 
+    both primary and replicas. When we enable replicas after the indexing, the 
+    serialized graphs are directly copied.
+ ```
+    
+* Increase number of indexing threads
+  ```
+    If the hardware we choose have multiple cores, we could allow multiple threads
```
jmazanec15: has instead of have.

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
...
+    If the hardware we choose have multiple cores, we could allow multiple threads 
+    in graph construction and there by speed up the indexing process. You could determine
+    the number of threads to be alloted by using the following setting   
+    https://github.com/opendistro-for-elasticsearch/k-NN#knnalgo_paramindex_thread_qty.
+     
+    Please keep an eye on CPU utilization and choose right number of threads. Since graph
+    construction is costly, having multiple threads can put additional load on CPU. 
+  ```
+    
+* Index all docs (Perform bulk indexing)
```
jmazanec15: Add comment here describing in more depth

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
...
+
+* Forcemerge
```
jmazanec15: I dont think you will need to do this. If forcemerge times out, it will continue to run.

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
...
+    
+    curl -X POST "localhost:9200/myindex/_forcemerge?max_num_segments=1&pretty"
+    
+* Call refresh 
+
+ Might not needed but to ensure the buffer is cleared and all segments are up. 
+ ```
+  POST /twitter/_refresh
+```
+* Add replicas (replica shards)
```
jmazanec15: Please add description for how this helps

PerformanceTuning.md
```diff
@@ -0,0 +1,147 @@
...
+* Index all docs (Perform bulk indexing)
+
+* Forcemerge 
+  
+ Forcemerge is a costly operation and could take a while depending on number of segments and size of the segments.
+ To ensure force merge is completed, we could keep calling forcemerge with 5 minute interval till you get 200 response.
+    
+    curl -X POST "localhost:9200/myindex/_forcemerge?max_num_segments=1&pretty"
+    
+* Call refresh
```
jmazanec15: Could you add description here?

PerformanceTuning.md
vamshin: Done

PerformanceTuning.md
vamshin: Done

PerformanceTuning.md
vamshin: Good point. Done

PerformanceTuning.md
vamshin: Done

PerformanceTuning.md
vamshin: done

PerformanceTuning.md
vamshin: I think its self explanatory.

PerformanceTuning.md
vamshin: Correct it will continue to run but to figure out when forcemerge is completed, we need to kind of poll it. 
We could either keep calling it or poll the ```tasks``` api.

PerformanceTuning.md
vamshin: Done

PerformanceTuning.md
vamshin: Done

PerformanceTuning.md
vamshin: I changed my mind to not include this step. Default settings for translog is 512MB which seems to be safe setting. 10GB would be too much for smaller instance types with low disk space .

PerformanceTuning.md
vamshin: Yes. I pointed the [doc](https://www.elastic.co/guide/en/elasticsearch/reference/master/tune-for-indexing-speed.html#_disable_replicas_for_initial_loads) to read more about this section. It is covered there. To avoid confusion adding a note in the paragraph.

PerformanceTuning.md
vamshin: Elasticsearch uses single thread for IndexSearcher and hence segments are searched sequentially. Parallelism in shard is not achieved with current Elasticsearch code.

PerformanceTuning.md
vamshin: Good point to call the trade offs. Added information in the respective sections.

PerformanceTuning.md
jmazanec15: Ah so I think whats confusing on this point is that it references "the following steps", but they are bullet points. So, its more about the workflow that should be followed as opposed to independent tips.

PerformanceTuning.md
jmazanec15: Is there documentation for this?

jmazanec15 (commented):
From [this](https://github.com/opendistro-for-elasticsearch/k-NN/blob/330076a7c161280ab679c26226c2f44ec1e19d5a/PerformanceTuning.md) it looks like all headers that do not have a space between them and the title do not show up as headers.

Still need to review more, will continue this week.

PerformanceTuning.md
vamshin: Correct this is more like a workflow of steps. Changed to numbers instead of bullet points.

PerformanceTuning.md
vamshin: I could not find a documentation but here is the code pointers. ```IndexSearcher``` class has couple of constructors and one of the [constructor](https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L196) accepts ```Executor``` class which runs the segments parallely by batching them as per the size. Elasticsearch uses other [constructor](https://github.com/apache/lucene-solr/blob/master/lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java#L183) which does not have ```Executor``` and segments are searched sequentially

PerformanceTuning.md
```diff
@@ -0,0 +1,157 @@
+#KNN Performance Tuning
```
jmazanec15: Add space between # and KNN

PerformanceTuning.md
```diff
@@ -0,0 +1,157 @@
...
+
+
+In this document we provide recommendations for performance tuning to improve indexing/search performance with the k-NN plugin.  From a high level k-NN works on following principles:
+
+* Graphs are created per (Lucene) segment
+* Queries execute on segments sequentially inside the shard (same as any other Elasticsearch query) 
+* Each graph in the segment returns *<=k* neighbors. 
+* Coordinator node picks up final *size* number of neighbors from the neighbors returned by each shard
+
+##Indexing Performance Tuning
```
jmazanec15: Add Space

PerformanceTuning.md
```diff
@@ -0,0 +1,157 @@
...
+* Index all docs (Perform bulk indexing)
+
+* Forcemerge 
+  
+ Forcemerge is a costly operation and could take a while depending on number of segments and size of the segments.
+ To ensure force merge is completed, we could keep calling forcemerge with 5 minute interval till you get 200 response.
+    
+    curl -X POST "localhost:9200/myindex/_forcemerge?max_num_segments=1&pretty"
+    
+* Call refresh
```
jmazanec15: If we already called force merge, we dont need to call refresh, correct?

PerformanceTuning.md
```diff
@@ -0,0 +1,157 @@
...
+   "knn": {
+    "v": {
+      "vector": [-0.16490704,-0.047262248,-0.078923926],
+      "k": 50
+     }       
+   }
+ }
+}
+```
+##Improving Recall
```
jmazanec15: Add space

PerformanceTuning.md
```diff
@@ -0,0 +1,157 @@
...
+
+Recall can be configured by adjusting the algorithm parameters of hnsw algorithm exposed through index settings. Algorithm params that control the recall are *m, ef_construction, ef_search*. For more details on influence of algorithm parameters on the indexing, search recall, please refer this  doc (https://github.com/nmslib/hnswlib/blob/master/ALGO_PARAMS.md).  Increasing these values could help recall(better search results) but at the cost of higher memory utilization and increased indexing time. Our default values work on a broader set of use cases from our experiments but we encourage users to run their own experiments on their data sets and choose the appropriate values. You could refer to these settings in this section (https://github.com/opendistro-for-elasticsearch/k-NN#index-level-settings). We will add details on our experiments shortly here.
+
+##Memory Estimation
```
jmazanec15: Add space

PerformanceTuning.md
```diff
@@ -0,0 +1,157 @@
...
+    * (8 * M) = 4 bytes/edge * 2 levels/node *  M edge/level
+        * Note — as an estimation, each node will have membership in roughly 2 layers, and, on each layer, it will have M edges
+    * 1.1 = an extra 10% buffer for other meta data in the data structure
+* Example:- Let us assume
+    * 1 Million vectors 
+    * 256 Dimensions (2^8)
+    * M = 16 (default setting of HNSW)
+        * Memory required for !M vectors = 1.1*(4*256 + 8*16) *1M Bytes =~ 1.26GB 
+
+##Monitoring
```
jmazanec15: Add space

PerformanceTuning.md
jmazanec15: Interesting, makes sense thanks

PerformanceTuning.md
```diff
@@ -0,0 +1,157 @@
...
+        {
+            "index" : {
+                "refresh_interval" : "1m"
+            }
+        }
+ ```
+
+Please refer following doc (https://www.elastic.co/guide/en/elasticsearch/reference/master/tune-for-indexing-speed.html) for more details on improving indexing performance in general.
+
+##Search Performance Tuning
```
jmazanec15: Add space

PerformanceTuning.md
```diff
@@ -0,0 +1,157 @@
...
+
+To improve Search performance it is necessary to keep the number of segments under control. Lucene's IndexSearcher will search over all of the segments in a shard to find the 'size' best results. But, because the complexity of search for the HNSW algorithm is logarithmic with respect to the number of vectors, searching over 5 graphs with a 100 vectors each and then taking the top size results from ```5*k``` results will take longer than searching over 1 graph with 500 vectors and then taking the top size results from k results.
```
jmazanec15, line 81: This needs a header like the other sections. Also, good explanation. I think adding a link to this newly created section from the force merge section would be helpful.

PerformanceTuning.md
vamshin: Done

PerformanceTuning.md
vamshin: Done

PerformanceTuning.md
vamshin: Correct. This step is just to ensure all documents are available for search. More like a safety check

PerformanceTuning.md
vamshin: Done.

PerformanceTuning.md
vamshin: Done

PerformanceTuning.md
vamshin: Done

PerformanceTuning.md
vamshin: Done

PerformanceTuning.md
vamshin: makes sense

PerformanceTuning.md
```diff
@@ -0,0 +1,154 @@
...
+
+### Warm up
+
+The graphs are constructed during indexing, but they are loaded into memory during the first search. The way search works in Lucene is that each segment is searched sequentially (so, for k-NN, each segment returns up to k nearest neighbors of the query point) and the top ```size``` number of results based on the score would be returned from all of the results returned by segements at a shard level(higher score --> better result). 
+
+Once a graph is loaded(graphs are loaded outside Elasticsearch JVM), we cache the graphs in memory. So the initial queries would be expensive in the order of few seconds and subsequent queries should be faster in the order of milliseconds(assuming knn circuit breaker is not hit).
+
+In order to avoid this latency penalty during your first queries, a user should use the warmup API on the indices they want to search. The API looks like this:
+
+GET /_opendistro/_knn/warmup/index1,index2,index3?pretty
```
jmazanec15, line 95: Minor: wrap in code block

PerformanceTuning.md
```diff
@@ -0,0 +1,154 @@
...
+* Graphs are created per (Lucene) segment
+* Queries execute on segments sequentially inside the shard (same as any other Elasticsearch query) 
+* Each graph in the segment returns *<=k* neighbors. 
+* Coordinator node picks up final *size* number of neighbors from the neighbors returned by each shard
+
+## Indexing Performance Tuning
+
+The following steps could help improve indexing performance especially when you plan to index large number of vectors at once. 
+
+1 Disable refresh interval  (Default = 1 sec)
```
jmazanec15: Minor: add "." after numbering

PerformanceTuning.md
vamshin: Done

PerformanceTuning.md
vamshin: Done

jmazanec15 (approved):
Looks good to me. Thanks for adding this @vamshin!

Issue #181, opened 2020-07-27
Update Guava to latest (29.0-jre)


jmazanec15:
We need to update Guava from 15.0 to the latest version (29.0-jre) to remove security vulnerability that can cause DOS, [CVE-2018-10237](https://nvd.nist.gov/vuln/detail/CVE-2018-10237).

From description. two classes are impacted:
* AtomicDoubleArray
* CompoundOrdering

We do not explicitly use either of these classes. Investigating impact on k-NN plugin.

PR #182, opened 2020-07-27
Update guava version to 29.0

jmazanec15:
*Issue #, if available:*
#181 

*Description of changes:*
Updated Guava from 15 to latest (29.0). Along the way, Guava removed error_prone_annotations and failureaccess from it, so we have to add those in order to prevent the build from breaking.


vamshin (approved):
LGTM! Thanks for the fix

Issue #178, opened 2020-07-25
KNNESSettingsTestIT failure in action "Build and Test k-NN Plugin"


jmazanec15:
KNNESSettingsTestIT fails on Ubuntu during [Github action](https://github.com/opendistro-for-elasticsearch/k-NN/runs/899654200?check_suite_focus=true). Additionally, was able to repro on Ubuntu 16:
```
com.amazon.opendistroforelasticsearch.knn.index.KNNESSettingsTestIT > testCacheRebuiltAfterUpdateIndexSettings FAILED
    java.lang.AssertionError: expected:<1> but was:<4>
        at __randomizedtesting.SeedInfo.seed([C11B19DAD31088ED:8E515F7F7E6EEC35]:0)
        at org.junit.Assert.fail(Assert.java:88)
        at org.junit.Assert.failNotEquals(Assert.java:834)
        at org.junit.Assert.assertEquals(Assert.java:645)
        at org.junit.Assert.assertEquals(Assert.java:631)
        at com.amazon.opendistroforelasticsearch.knn.index.KNNESSettingsTestIT.testCacheRebuiltAfterUpdateIndexSettings(KNNESSettingsTestIT.java:130)
```

The code that was being checked in was #170 , however, this does not impact any testing logic or that action logic. So, this is most likely a flaky test. I think it may be related to the tests in #162, even though the action for that PR [did not fail](https://github.com/opendistro-for-elasticsearch/k-NN/actions/runs/177579646).

jmazanec15:
So, I think the issue is that the indices from another test are deleted, but not cleared from the cache, before the next test begins.

Indices for rest test cases are deleted after a test [here](https://github.com/elastic/elasticsearch/blob/dd67392b66384e1dda3c8ac77c6263cd1cc204dc/test/framework/src/main/java/org/elasticsearch/test/rest/ESRestTestCase.java#L308)

jmazanec15:
For this, the cache should be cleared before the next test case starts. We can implement this by adding on to `cleanUpCluster`. When we have the clear cache API, we should use that to accomplish this. However, until then, we will need to use a workaround. One way to do this is to bounce one k-NN setting. This will result in an empty cache. @vamshin does this sound good?

vamshin:
@jmazanec15 LGTM! 
Lets leave a TODO comment on the test case and tag the issue to track the fix with clear cache api.

jmazanec15:
[...]

PR #183, opened 2020-07-27
Flaky rest test case fix

jmazanec15:
*Issue #, if available:*
#178 

*Description of changes:*
Introduce clearing of the cache after each rest test case runs. Currently, we do this by bouncing a random setting. However, in the future, we should replace this with the clear cache API.

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java
@@ -30,8 +30,8 @@ import org.elasticsearch.common.xcontent.XContentFactory;
 import org.elasticsearch.common.xcontent.XContentType;
 import org.elasticsearch.rest.RestStatus;
 import org.elasticsearch.test.rest.ESRestTestCase;
-import org.junit.After;
 import org.junit.AfterClass;
+import org.junit.Before;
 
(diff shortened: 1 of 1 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java
```diff
@@ -413,4 +420,16 @@ protected String getIndexSettingByName(String indexName, String settingName) thr
...
         return (String)settings.get(settingName);
     }
+
+    /**
+     * Clear cache
+     *
+     * This function is a temporary workaround. Right now, we do not have a way of clearing the cache except by deleting
+     * an index or changing k-NN settings. That being said, this function bounces a random k-NN setting in order to
+     * clear the cache.
+     */
```
vamshin, line 430: Thanks for the comment here. Please create an issue so that it would be easy to track and get back

jmazanec15: #68 exists for the flush API. Should I create another for this? Or add comment to that issue?

vamshin: Adding a comment to the existing issue should be fine.

vamshin (approved):
LGTM!

PR #186, opened 2020-07-29
Fix cluster setting example in README.md

dengxianjie:
number 15 will throw "reason" : "failed to parse setting [knn.cache.item.expiry.minutes] with value [15] as a time value: unit is missing or unrecognized"
and "15m" works!

*Issue #, if available:*

*Description of changes:*


vamshin (approved):
LGTM!  Good catch and thanks for the fix

jmazanec15 (approved):
Oh nice good catch @dengxianjie !

Issue #169, opened 2020-07-17
Use containers for library artifact build


jmazanec15:
We should use containers for the library build and release in our Github Action to control the environment better. 

Additionally, for the DEB artifact, we should continue to use Ubuntu 16.04. But for RPM, we should use Centos 7 as opposed to Ubuntu.

peterzhuamazon:
As we discussed @jmazanec15 since CentOS have the lowest version of glibc, it would make sense to compile all distros of knnlib (zip/deb/rpm) in CentOS.

On Debian Stretch (9):
ldd (Debian GLIBC 2.24-11+deb9u4) 2.24

On CentOS 7:
ldd (GNU libc) 2.17

On a side note, Debian Jessie (8) is out of the LTS since June 2020, so we are not supporting it anymore.
https://wiki.debian.org/LTS

PR #194, opened 2020-08-06
Add AWS CLI in order to ship library artifacts from container

jmazanec15:
*Issue #, if available:*
#169 

*Description of changes:*
Because we now build and ship library artifacts from Centos container, we need to manually install the AWS CLI in the container. Verified that this installation works on Centos 7.

Related to test Github action failure: https://github.com/opendistro-for-elasticsearch/k-NN/runs/951599939?check_suite_focus=true

Documentation for installing AWS CLI on linux: https://docs.aws.amazon.com/cli/latest/userguide/install-cliv2-linux.html

### DIFF
--- a/.github/workflows/CD.yml
+++ b/.github/workflows/CD.yml
@@ -70,9 +70,13 @@ jobs:
       - name: Install dependencies
         run: |
           yum update -y
-          yum install -y cmake rpm-build gcc-c++ make epel-release
+          yum install -y cmake rpm-build gcc-c++ make epel-release unzip
           yum repolist
           yum install -y dpkg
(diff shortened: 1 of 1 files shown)


vamshin:
Why do we need AWS cli?

jmazanec15:
@vamshin we need it because we use it to copy the built artifacts to S3 [here](https://github.com/opendistro-for-elasticsearch/k-NN/pull/194/files#diff-07e90143318691bf799f65ea1463bc2dR102). The Centos container used for building the artifacts does not come with it, like the Github Actions Ubuntu runner.

vamshin (approved):

Issue #50, opened 2020-02-14
Support custom scoring function for vectors. Using k-NN scores in a script_score query


mikepalei:
Hi guys, is it possible to use your plugin in a script_score query (like below). If it is, could you please provide an example?

```
GET my_index/_search
{
  "query": {
    "script_score": {
      "query" : {
        "bool" : {
          "filter" : {
            "term" : {
              "status" : "published" 
            }
          }
        }
      },
      "script": {
        "source": "cosineSimilarity(params.query_vector, 'my_dense_vector') + 1.0", 
        "params": {
          "query_vector": [4, 3.4, -0.2]  
        }
      }
    }
  }
}
```

vamshin:
Hi @mikepalei ,

We do not have support for painless query.

mikepalei:
So there is no way to run your plugin against a limited set of vectors and not on the entire index?

vamshin:
Hi @mikepalei,

Correct. At this point we do not have a way to run dynamically on the limited set of vectors.
Tagging this issue as Feature request. Feel free to +1.

dwang-sflscientific:
I saw the PR is merged. Is the feature ready to use?

vamshin:
Hi @dwang-sflscientific, this feature would be available in the next release i.e ODFE 1.11.

PR #196, opened 2020-08-20
Pre filter support through custom scoring

vamshin:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/k-NN/issues/50

- As of today k-NN search queries in combination with filters/boolean queries always work as a post filter search.  k-Nearest Neighbors are evaluated first and then other queries are run on top of the returned k results . When k is small, it is possible we end up less than required k Neighbors after applying the filter.
- This issue becomes more evident especially when k-NN search is combined with filter queries targeting smaller subset of vectors in the large data set.

*Description of changes:*
We define our own custom Script Engine which uses the language name “knn”.  It implements a script class called *VectorScoreScript* which would be called to override each document’s score with the chosen similarity space like Euclidean, Cosine, Dot product etc. Customer could invoke the script using the “script_score” functionality by passing the language name as “knn” and required parameters. 

* The script_score functionality can be incorporated by extending ‘*ScriptPlugin*’. 
* *script_score* is evaluated for the filtered documents at segment level inside a shard. 
* Customers would use *size* attribute to define the ‘k’.
* script would be compiled once on the first request and cached for later 


vamshin:
This Pr focus on adding support for Euclidean(l2).  Follow up Prs would have 

- cosine similarity support.
- metrics for this feature
- test cases

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/util/KNNConstants.java
```diff
@@ -21,4 +21,6 @@
     public static final String HNSW_ALGO_EF_CONSTRUCTION = "efConstruction";
     public static final String HNSW_ALGO_EF_SEARCH = "efSearch";
     public static final String HNSW_ALGO_INDEX_THREAD_QTY = "indexThreadQty";
+    public static final String L2 = "l2";
+    public static final String COSINESIMIL = "cosinesimil";
```
jmazanec15, line 25: If we do not support Cosine Similarity yet, remove from this PR

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/VectorScoreScript.java
```diff
@@ -0,0 +1,149 @@
...
+            if (bytesref == null) {
+                return Float.MIN_VALUE;
+            }
+            try (ByteArrayInputStream byteStream = new ByteArrayInputStream(bytesref.bytes, bytesref.offset, bytesref.length);
+                 ObjectInputStream objectStream = new ObjectInputStream(byteStream)) {
+                doc_vector = (float[]) objectStream.readObject();
+            } catch (ClassNotFoundException e) {
+                throw new RuntimeException(e);
+            }
+            if (KNNConstants.L2.equalsIgnoreCase(similaritySpace)) {
```
jmazanec15: Where do we check that the input vector has the correct dimensions?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/VectorScoreScript.java
```diff
@@ -0,0 +1,149 @@
...
+     *                          it into an {@link org.apache.lucene.search.Explanation}
+     * @return score of the vector to the query vector
+     */
+    @Override
+    public double execute(ScoreScript.ExplanationHolder explanationHolder) {
+        float score = Float.MIN_VALUE;
+        try {
+            float[] doc_vector;
+            BytesRef bytesref = binaryDocValuesReader.binaryValue();
+            // If there is no vector for the corresponding doc then it should be not considered for nearest
```
jmazanec15: In what case does this happen?

jmazanec15 (commented):
I think a couple todos for this PR:
1. Add usage information in README, including limitations
2. Add unit and integration tests

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringScriptEngine.java
```diff
@@ -0,0 +1,40 @@
...
+    private static final String SCRIPT_SOURCE = "knn_score";
+
+    @Override
+    public String getType() {
+        return NAME;
+    }
+
+    @Override
+    public <FactoryType> FactoryType compile(String name, String code, ScriptContext<FactoryType> context, Map<String, String> params) {
+        if (context.equals(ScoreScript.CONTEXT) == false) {
```
wnbts: minor. typically java developers do constant::equals to avoid npe, like line 29.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/VectorScoreScript.java
```diff
@@ -0,0 +1,149 @@
...
+ * on a per document basis.
+ */
+public class VectorScoreScript extends ScoreScript {
+
+    private BinaryDocValues binaryDocValuesReader;
+    private final float[] inputVector;
+    private final String similaritySpace;
+
+    public float l2Squared(float[] queryVector, float[] inputVector) {
+        long squaredDistance = 0;
```
wnbts: question. why is the type not float?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/VectorScoreScript.java
```diff
@@ -0,0 +1,149 @@
...
+        for (int i = 0; i < inputVector.length; i++) {
+            squaredDistance += Math.pow(queryVector[i]-inputVector[i], 2);
```
wnbts: this implementation might not be the most efficient due to type casting and power.

```
float diff = queryVector[i] - inputVector[i];
squaredDistance += diff * diff;
```
should be 2x faster

```
float diff = queryVector[i] - inputVector[i];
squaredDistance = Math.fma(diff, diff, squaredDistance);
```
should also be fast but not as much as the one above

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/VectorScoreScript.java
```diff
@@ -0,0 +1,149 @@
...
+        final Object vector = params.get("vector");
+        this.similaritySpace = similaritySpace;
+        if(vector != null) {
+            final ArrayList<Double> tmp = (ArrayList<Double>) vector;
+            inputVector = new float[tmp.size()];
+            for (int i = 0; i < inputVector.length; i++) {
+                inputVector[i] = tmp.get(i).floatValue();
+            }
+        } else {
+            inputVector = null;
```
wnbts: question. should this case error out?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -13,6 +13,25 @@ public static float l2Squared(float[] queryVector, float[] inputVector) {
         return squaredDistance;
     }
 
+    public static float cosinesimilOptimized(float[] queryVector, float[] inputVector, double normQueryVector) {
+        double dotProduct = 0.0f;
+        double normInputVector = 0.0f;
```
wnbts: since input and outputs are all float, why not using float?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -13,6 +13,25 @@ public static float l2Squared(float[] queryVector, float[] inputVector) {
...
+        if (normQueryVector == -1) {
```
wnbts: this check should preferably be avoided for every input vector

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -0,0 +1,74 @@
...
+            throw new IllegalStateException("Normalized query vector cannot be negative");
+        }
+        for (int i = 0; i < queryVector.length; i++) {
+            dotProduct += queryVector[i] * inputVector[i];
+            normInputVector += inputVector[i] * inputVector[i];
+        }
+        // Divide by zero check
+        double normalizedProduct = normQueryVector * normInputVector;
+        if (normalizedProduct == 0 ) {
+            return Float.MIN_VALUE;
```
wnbts: this check can be saved for most vectors by try-catching the unlikely case from the division below

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -0,0 +1,74 @@
...
+
+    public static double getVectorMagnitudeSquared(float[] inputVector) {
+        if (null == inputVector) {
+            throw new IllegalStateException("vector magnitude cannot be evaluated as it is null");
+        }
+        double normInputVector = 0.0f;
+        for (int i = 0; i < inputVector.length; i++) {
+            normInputVector += inputVector[i] * inputVector[i];
+        }
+        return (float) normInputVector;
```
wnbts: the return type should be float

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNVectorScoreScript.java
```diff
@@ -0,0 +1,166 @@
...
+ * on a per document basis.
+ */
+public class KNNVectorScoreScript extends ScoreScript {
+
+    private BinaryDocValues binaryDocValuesReader;
+    private final float[] queryVector;
+    private final String similaritySpace;
+    private double queryVectorSquaredMagnitude = -1;
+
+    public float l2Squared(float[] queryVector, float[] inputVector) {
```
wnbts: this method is not used

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/util/KNNConstants.java
vamshin: Was adding incrementally in the PR should be available now.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/VectorScoreScript.java
vamshin: Added part of the execute() function

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/VectorScoreScript.java
vamshin: If for some reason, some documents do not have vector field then we need to handle this case gracefully.
I am doing some more tests around this.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringScriptEngine.java
vamshin: Fixed.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/VectorScoreScript.java
vamshin: Should be float. Corrected. thanks

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/VectorScoreScript.java
vamshin: Fixed. thanks

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: Math.sqrt() function requires double. So doing all the manipulations in double and type cast to float later

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: This case would not occur from Script. Since this function is part of ```KNNScoringUtil``` Util class, wanted to add this check to ensure consumers of this function take care of this value.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: Makes sense. Fixed. Thanks

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: since this value would be used in Math.sqrt function, wanted to keep it as double.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNVectorScoreScript.java
vamshin: Correct. Removed.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
wnbts: note that every casting, even implicit, takes away resource and gives no gains

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
wnbts: can this check be moved to be done exactly once at the factory when the script is created?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
wnbts: if the return must be double, then the casting to float at line is unnecessary

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
NelsonBurton: Just to add on, I ran some benchmarking on a single i3.4xl of a tight dot product loop, and found the casting back and forth to double slowed down the calculation measurably.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    public static float cosinesimil(float[] queryVector, float[] inputVector) {
+        double dotProduct = 0.0f;
+        double normQueryVector = 0.0f;
+        double normInputVector = 0.0f;
+        for (int i = 0; i < queryVector.length; i++) {
+            dotProduct += queryVector[i] * inputVector[i];
+            normQueryVector += queryVector[i] * queryVector[i];
+            normInputVector += inputVector[i] * inputVector[i];
```
NelsonBurton, line 66: Recomputing the query vector of every invocation will add unnecessary computation, as it can be computed once and passed it, like in the function above

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/VectorScoreScript.java
NelsonBurton: Just out of curiosity, did you performance test the Math.fma computation as compared to the += diff * diff on an Intel CPU? This article implies that fma could be more performant as it can use a special intel instruction set: https://software.seek.intel.com/boosting-java-performance

wnbts: great ask!

I did test Math.fma and the performance appeared consistently slower than += on intel cpu with avx/avx2. It could be the overhead from underlying extra if, assert, double casting. More test results are always better.

Also note that on earlier cpu models, the fma implementation might be even worse. see https://stackoverflow.com/questions/44808081/what-are-the-accuracy-performance-benefits-of-using-math-fma

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: Good point. Fixed

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: Removed this check as this case does not occur

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: I added two versions of calculating cosine similarity score. This is unoptimized version of cosine similarity function which would be exposed to customer to call the function directly.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -0,0 +1,76 @@
...
+            return (float) (dotProduct / (Math.sqrt(normalizedProduct)));
+        } catch(ArithmeticException ex) {
+            logger.debug("Possibly Division by Zero Exception. Returning min score to put this result to end. " +
+                    "Current normalized product " + normalizedProduct);
+            return Float.MIN_VALUE;
+        }
+    }
+
+    public static float cosinesimil(float[] queryVector, float[] inputVector) {
+        double dotProduct = 0.0f;
```
wnbts: even this method is not used by the plugin, the implementation can be similar to the optimized version. unnecessary casting from float to double can be removed.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNVectorScoreScript.java
```diff
@@ -0,0 +1,162 @@
...
+        super(params, lookup, leafContext);
+        // get query vector - convert to primitive
+        final Object vector = params.get("vector");
+        this.similaritySpace = similaritySpace;
+        this.queryVector = queryVector;
+        this.queryVectorSquaredMagnitude = queryVectorSquaredMagnitude;
+        try {
+            this.binaryDocValuesReader = leafContext.reader().getBinaryDocValues(field);
+            if(this.binaryDocValuesReader == null) {
+                throw new IllegalStateException();
```
wnbts: minor. an error message is clearer.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNVectorScoreScript.java
```diff
@@ -0,0 +1,162 @@
...
+            this.similaritySpace = space != null? (String)space: KNNConstants.L2;
+            this.qVector = KNNScoringUtil.convertVectorToPrimitive(params.get("vector"));
+            // Optimization for cosinesimil
+            if (KNNConstants.COSINESIMIL.equalsIgnoreCase(similaritySpace)) {
+                // calculate the magnitude
+                qVectorSquaredMagnitude = KNNScoringUtil.getVectorMagnitudeSquared(qVector);
+            }
+        }
+
+        private void validateParams(Map<String, Object> params) {
```
wnbts: minor. space can also use validation.

wnbts (approved):
unit test can help verify correctness

jmazanec15:
A couple comments:
1. During search, we should validate the space type and throw error if invalid one is provided
2. I think there may be a little confusion with spaces. Say someone index's the data with "space_type" set to "cosinesimil". Then they go to search and they do not specify the "space" in the search. This will default to "l2". I think instead, it should default to the space_type specified during indexing.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: This method would be used in future part of https://github.com/opendistro-for-elasticsearch/k-NN/issues/213.  Fixed the typecasting issue.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNVectorScoreScript.java
vamshin: Fixed. thanks

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNVectorScoreScript.java
vamshin: Added Validation for space.

vamshin:
> A couple comments:
> 
> 1. During search, we should validate the space type and throw error if invalid one is provided
> 2. I think there may be a little confusion with spaces. Say someone index's the data with "space_type" set to "cosinesimil". Then they go to search and they do not specify the "space" in the search. This will default to "l2". I think instead, it should default to the space_type specified during indexing.

1. Good point. Fixed 1.
2. Good point. Removed default. User has to specify the space type in the script. Agree its better to let user assert by mentioning space type.

vamshin:
> I think a couple todos for this PR:
> 
> 1. Add usage information in README, including limitations
> 2. Add unit and integration tests


Created issue for writing detailed README https://github.com/opendistro-for-elasticsearch/k-NN/issues/222

wnbts (approved):

jmazanec15 (approved):

vamshin:
I will keep this PR open to fix code coverage. There is a 0.9% drop in coverage with this feature

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringScriptEngine.java
```diff
@@ -0,0 +1,40 @@
...
+    public static final String SCRIPT_SOURCE = "knn_score";
+
+    @Override
+    public String getType() {
+        return NAME;
+    }
+
+    @Override
+    public <FactoryType> FactoryType compile(String name, String code, ScriptContext<FactoryType> context, Map<String, String> params) {
+        if (ScoreScript.CONTEXT.equals(context) == false) {
```
ylwu-amzn: How about simplify as `if (!ScoreScript.CONTEXT.equals(context))` ?

Is it possible that KNN script will have invalid context ? I see we have validated the script source in line 29. We need line 25 as the script context is possibly wrong?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringScriptEngine.java
```diff
@@ -0,0 +1,40 @@
...
+        if (!SCRIPT_SOURCE.equals(code)) {
+            throw new IllegalArgumentException("Unknown script name " + code);
+        }
+        ScoreScript.Factory factory = KNNVectorScoreScript.VectorScoreScriptFactory::new;
+        return context.factoryClazz.cast(factory);
+    }
+
+    @Override
+    public Set<ScriptContext<?>> getSupportedContexts() {
+        return null;
```
ylwu-amzn, line 38: Why return null here? We don't need context for KNN score script?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -0,0 +1,76 @@
...
+        float dotProduct = 0.0f;
+        float normInputVector = 0.0f;
+        for (int i = 0; i < queryVector.length; i++) {
+            dotProduct += queryVector[i] * inputVector[i];
+            normInputVector += inputVector[i] * inputVector[i];
+        }
+        float normalizedProduct = normQueryVector * normInputVector;
+        try {
+            return (float) (dotProduct / (Math.sqrt(normalizedProduct)));
+        } catch(ArithmeticException ex) {
```
ylwu-amzn: How about we use if/else to make code cleaner like the method below line47?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -0,0 +1,76 @@
...
+    public static float l2Squared(float[] queryVector, float[] inputVector) {
+        float squaredDistance = 0;
+        for (int i = 0; i < inputVector.length; i++) {
+            float diff = queryVector[i]-inputVector[i];
+            squaredDistance += diff * diff;
+        }
+        return squaredDistance;
+    }
+
+    public static float cosinesimilOptimized(float[] queryVector, float[] inputVector, float normQueryVector) {
```
ylwu-amzn, line 37: Seems this method is mostly the same with the method below. Can we unify them by abstracting the common part?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    @SuppressWarnings("unchecked")
+    public static float[] convertVectorToPrimitive(Object vector) {
+        float[] primitiveVector = null;
+        if(vector != null) {
+            final ArrayList<Double> tmp = (ArrayList<Double>) vector;
+            primitiveVector = new float[tmp.size()];
+            for (int i = 0; i < primitiveVector.length; i++) {
+                primitiveVector[i] = tmp.get(i).floatValue();
```
ylwu-amzn, line 89: Is it possible that `tmp.get(i)` is null?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNVectorScoreScript.java
```diff
@@ -0,0 +1,166 @@
...
+    /**
+     * This function called for each doc in the segment. We evaluate the score of the vector in the doc
+     *
+     * @param explanationHolder A helper to take in an explanation from a script and turn
+     *                          it into an {@link org.apache.lucene.search.Explanation}
+     * @return score of the vector to the query vector
+     */
+    @Override
+    public double execute(ScoreScript.ExplanationHolder explanationHolder) {
+        float score = Float.MIN_VALUE;
```
ylwu-amzn, line 37: Why use float while this function returns double?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNVectorScoreScript.java
```diff
@@ -0,0 +1,166 @@
...
+                throw new IllegalStateException("[KNN] query vector and field vector dimensions mismatch. " +
+                        "query vector: " + queryVector.length + ", stored vector: " + doc_vector.length);
+            }
+
+            if (KNNConstants.L2.equalsIgnoreCase(similaritySpace)) {
+                score = KNNScoringUtil.l2Squared(this.queryVector, doc_vector);
+                score = 1/(1 + score);
+            } else if (KNNConstants.COSINESIMIL.equalsIgnoreCase(similaritySpace)) {
+                // Scores cannot be negative so add +1 to the cosine score
+                score = 1 + KNNScoringUtil.cosinesimilOptimized(this.queryVector, doc_vector, this.queryVectorSquaredMagnitude);
```
ylwu-amzn, line 63: From https://en.wikipedia.org/wiki/Cosine_similarity,  seems others generally use `1 - cosine similarity`, any special logic here we use `1 + cosine similarity` ?
If just make the score positive, we can use `1 - cosine similarity` as well.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: No. It cannot be null.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNVectorScoreScript.java
vamshin: All the scores are computed in Float. To keep it consistent returning ```Float.MIN```.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNVectorScoreScript.java
vamshin: We cannot use ```1 - cosine similarity```. This would invert the scores and return wrong results.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: We would like to avoid additional checks where ever possible to improve performance. So ```try catch``` takes care of edge cases.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: I think it makes sense to keep it this way.  Clean to understand.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringScriptEngine.java
vamshin: There is no context for KNN score script.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringScriptEngine.java
vamshin: As mentioned in the log message ```KNN Vector scoring scripts cannot be used for context ```.  

Will fix ```if (!ScoreScript.CONTEXT.equals(context))```

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNVectorScoreScript.java
ylwu-amzn: Make sense. So the greater score means it's more relevant?

vamshin: Correct.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
ylwu-amzn: Java will throw `ArithmeticException` for integer division by zero, but not throw `ArithmeticException` for float division by zero, check this [stack overflow question](https://stackoverflow.com/questions/12954193/why-does-division-by-zero-with-floating-point-or-double-precision-numbers-not), will return `Infinity` for `float/0`,  so we can't catch `ArithmeticException` and return `Float.MIN_VALUE`

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -0,0 +1,76 @@
...
+        if (normalizedProduct == 0 ) {
+            return Float.MIN_VALUE;
+        }
+        return (float) (dotProduct / (Math.sqrt(normalizedProduct)));
+    }
+
+    @SuppressWarnings("unchecked")
+    public static float[] convertVectorToPrimitive(Object vector) {
+        float[] primitiveVector = null;
+        if(vector != null) {
```
ylwu-amzn: Minor, missing space after `if`, suggest to format the code.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
ylwu-amzn: If we need to make some change in future, we may need to change both methods. But seems it's not high possible that we will change these two methods as they contain just simple calculation. I'm ok to keep it as is. 
How about we add some comments to these methods to explain differences and the best use cases ? I think it will help others to understand.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: Fixed.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: Sure added comments

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
vamshin: Good catch. Able to catch this in unit tests. Added ```if``` condition to check for division by zero issue. Adding ```if``` check could have minor performance impact. Will keep an eye in performance tests

ylwu-amzn: Thanks, looks good to me.

ylwu-amzn (approved):
LGTM. Thanks for the change.

williamdias:
@vamshin , do you know if the latest AWS Elasticsearch Service version supports knn custom scoring? If not, do you have an estimate of when it will be available?

vamshin:
> @vamshin , do you know if the latest AWS Elasticsearch Service version supports knn custom scoring? If not, do you have an estimate of when it will be available?

@williamdias , in AWS Elasticsearch service, custom scoring should be available from ES version 7.8 onwards.

Issue #198, opened 2020-08-25
ODFE 1.10 support with Elasticsearch OSS 7.9


vamshin:
Support for k-NN plugin for ODFE 1.10 version which runs on Elasticsearch OSS 7.9.0 version

PR #199, opened 2020-08-25
ODFE 1.10 support for k-NN plugin

vamshin:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/k-NN/issues/198

*Description of changes:*
Support for k-NN plugin for ODFE 1.10 version which runs on Elasticsearch OSS 7.9.0 version


build.gradle
```diff
@@ -102,7 +102,6 @@ task release(type: Copy, group: 'build') {
 //****************************************************************************/
 dependencies {
     compileOnly "org.elasticsearch:elasticsearch:${es_version}"
-    compileOnly group: 'com.google.errorprone', name: 'error_prone_annotations', version:'2.4.0'
```
jmazanec15, line 105: Why was this removed?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -56,38 +57,48 @@
...
         public static final String IGNORE_MALFORMED = "ignore_malformed";
     }
 
+    public KNNVectorFieldMapper clone() {
+        return (KNNVectorFieldMapper) super.clone();
+    }
+
     public static class Defaults {
         public static final Explicit<Boolean> IGNORE_MALFORMED = new Explicit<>(false, false);
-        public static final KNNVectorFieldType FIELD_TYPE = new KNNVectorFieldType();
```
jmazanec15, line 72: Why was KNNVectorField removed?

jmazanec15 (commented):
Changes should not be made to `jni/external/nmslib`.

On line 235 of build.graddle, could you update `https://opendistro.github.io/elasticsearch/downloads` to `https://opendistro.github.io/for-elasticsearch/downloads.html`

This line needs to be updated as well: https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/jni/CMakeLists.txt#L75.

build.gradle
vamshin: Elasticsearch pulls this dependency now in this release and there was a conflict, so i removed it to use Elasticsearch's dependency

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
vamshin: There is a refactoring around MappedFieldType and FieldType separating indexing and search attributes. More details here https://github.com/elastic/elasticsearch/pull/58160. As part of this i need to replace the MappedFieldType to FieldType for providing the indexing attributes.

vamshin:
> Changes should not be made to `jni/external/nmslib`.
> 
> On line 235 of build.graddle, could you update `https://opendistro.github.io/elasticsearch/downloads` to `https://opendistro.github.io/for-elasticsearch/downloads.html`
> 
> This line needs to be updated as well: https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/jni/CMakeLists.txt#L75.

Good catch. Fixed

jmazanec15 (approved):
LGTM

Issue #204, opened 2020-09-01
Restoring from ODFE < 1.7 snapshot failing when upgrading to ODFE 1.7 and above


jmazanec15:
In ODFE 1.7, we introduced the cosine similarity feature. To do this, we added an index setting called `index.knn.space_type`, which can be either `l2` or `cosinesimil`. Restoring a snapshot from a previous ODFE version to a version >= 1.7 will cause cluster to go red, with an error that looks like this:
```
{"error":{"root_cause":[{"type":"snapshot_restore_exception","reason":"[<snap rep0>:<snap name>/<id>] cannot restore index [<index>] because it cannot be upgraded"}],"type":"snapshot_restore_exception","reason":"[<snap repo>:<snap name>/<id>] cannot restore index [<index>] because it cannot be upgraded","caused_by":{"type":"illegal_state_exception","reason":"unable to upgrade the mappings for the index [[<index>/<id>]]","caused_by":{"type":"mapper_parsing_exception","reason":"Failed to parse mapping [_doc]: setting
 index.knn.space_type has not been registered","caused_by":{"type":"illegal_argument_exception","reason":"setting index.knn.space_type has not been registered"}}}},"status":500}
```

This seems to fail in KNNVectorFieldMapper, [here](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java#L147)

PR #205, opened 2020-09-01
Add default index settings when parsing index

jmazanec15:
*Issue #, if available:*
#204 

*Description of changes:*
Adds in default settings if an exception is thrown when attempting to get settings for the index during parsing. More details on why this is needed in the issue.


vamshin (approved):
LGTM! Thanks for the fix

Issue #210, opened 2020-09-02
Forcemerge causes NPE when documents with non vector fields get updated with vector fields


vamshin:
https://discuss.opendistrocommunity.dev/t/indexing-causing-index-to-enter-red-state/3490/9

jmazanec15:
To reproduce this bug:

1. Create knn index:
```
curl -X PUT "${HOST_NAME}/test_index" -H 'Content-Type: application/json' -d'
{
  "settings" : {
    "number_of_shards" :   1,
    "number_of_replicas" : 0,
    "index": {
        "knn": true
    }
  },
  "mappings": {
      "properties": {
        "knn_field": {
          "type": "knn_vector",
          "dimension": 2
        }
      }
  }
}
'
```
2. Index a document without knn field:
```
curl -X PUT "${HOST_NAME}/test_index/_doc/1" -H 'Content-Type: application/json' -d'
{
  "abc" : 1
}
'
```

3. Refresh index
```
curl ${HOST_NAME}/test_index/_refresh?pretty
```

4. update the document to have knn field
```
curl -X PUT "${HOST_NAME}/test_index/_doc/1" -H 'Content-Type: application/json' -d'
{
  "knn_field" : [11.5, 12.5]
}
'
```

5. Refresh index
```
curl ${HOST_NAME}/test_index/_refresh?pretty
```

6. Attempt to forcemerge
```
curl -XPOST ${HOST_NAME}/test_index/_forcemerge?max_num_segments=1
```

PR #212, opened 2020-09-04
NPE in force merge when non knn doc gets updated to knn doc across segments

vamshin:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/k-NN/issues/210

*Description of changes:*
NPE occurs forcemerge when non knn doc gets updated to knn doc. This happens when a a doc in non knn in one segment and gets updated to knn doc in another segment and forcemerge occurs.


jmazanec15 (approved):
Thanks for the fix!

Issue #203, opened 2020-08-28
confuse about cluster setting "cache_capacity_reached" and "knn.circuit_breaker.unset.percentage"


dengxianjie:
For deeply understand the behavior of the cache , I try some practice of cache setting likes below
```
PUT /_cluster/settings
{
    "persistent" : {
        "knn.plugin.enabled" : true,
        "knn.algo_param.index_thread_qty" : 20,
        "knn.cache.item.expiry.enabled": true,
        "knn.cache.item.expiry.minutes": "120m",
        "knn.memory.circuit_breaker.enabled" : true,
        "knn.memory.circuit_breaker.limit" : "10%",
        "knn.circuit_breaker.unset.percentage": 65
    }
}
```

And after indexing some vector into different data, the knn statistics api shows below. 

What makes me confused is that "graph_memory_usage_percentage" has been over 100, but "eviction_count" doesn't work(it always be 0) after I warmup some new index. 
The "cache_capacity_reached" should be true by the design but unfortunately it was still false even though "knn.circuit_breaker.unset.percentage" has totally reached  . Is it anything wrong?

My device is 250GB memory, JVM uses 31G.

```{
  "_nodes" : {
    "total" : 3,
    "successful" : 3,
    "failed" : 0
  },
  "cluster_name" : "odfe-cluster",
  "circuit_breaker_triggered" : false,
  "nodes" : {
    "xfvHY81rQjSQz5Wa0g8rmg" : {
      "miss_count" : 583,
      "graph_memory_usage_percentage" : 108.94683,
      "graph_query_requests" : 11629745,
      "graph_memory_usage" : 25204287,
      "cache_capacity_reached" : false,
      "graph_index_requests" : 57380,
      "load_exception_count" : 0,
      "load_success_count" : 544,
      "eviction_count" : 0,
      "indices_in_cache" : {
        "knn_pid_2020082711" : {
          "graph_memory_usage" : 7444895,
          "graph_memory_usage_percentage" : 32.180943,
          "graph_count" : 61
        },
        "knn_pid_2020082810" : {
          "graph_memory_usage" : 2702801,
          "graph_memory_usage_percentage" : 11.682997,
          "graph_count" : 29
        },
        "knn_pid_2020082812" : {
          "graph_memory_usage" : 2705674,
          "graph_memory_usage_percentage" : 11.6954155,
          "graph_count" : 29
        },
        "knn_pid_2020082801" : {
[...]

PR #215, opened 2020-09-04
Fix casting issue with cache expiration

jmazanec15:
*Issue #, if available:*
#203 

*Description of changes:*
Previously, we did not cast `KNNSettings.state().getSettingValue(KNNSettings.KNN_CACHE_ITEM_EXPIRY_TIME_MINUTES))` to `TimeValue` and then get the long value. This was causing the cache to never be rebuilt, causing the circuit breaker limit to exceed 100%. This change fixes that and adds a test to validate it.


vamshin (approved):
Good catch. LGTM!

Issue #197, opened 2020-08-21
existsQuery knn vecter is not run！


whitezdm:
when i use existquery to query the field definition knn vector,it not work . the item has date ,but when existsquery not find

vamshin:
Hi @whitezdm,

Could you share the example query?

whitezdm:
> Hi @whitezdm,
> 
> Could you share the example query?

{
    "query": {
        "exists": {
          "field": "imageVector"
        }
    }
}
imageVector item is vecter item 。i want find this item is not null. i use exits query in es 7.8 ,but it not work。i not know why？thanks

dnwarnock:
I've noticed the same thing. Outside the KNN query DSL the vector fields don't seem compatible with other ES query types.

vamshin:
@whitezdm @dnwarnock we are considering this as a feature request and we would update the availability of this feature on this thread.

whitezdm:
> @whitezdm @dnwarnock we are considering this as a feature request and we would update the availability of this feature on this thread.

very expect

jmazanec15:
@dnwarnock @whitezdm are there other ES query types that would be useful for your use cases?

whitezdm:
> @dnwarnock @whitezdm are there other ES query types that would be useful for your use cases?

other types can work very well.

dnwarnock:
It would be nice to be able to query for documents that have non null vector fields to collect metrics around how many documents for a given index have associated embeddings. Also the ability to rerank based off some other document feature/metadata would be useful.

hjander:
I am missing exactly this functionality. When will it be available in OpenDistro ? I am using AWS ES 7..7

jmazanec15:
Hi @hjander 

In opendistro, this will be available in opendistro-1.11, which will be available in the next couple of weeks.

At the moment, it is not yet determined when this will be in Amazon Elasticsearch Service.

We will update this thread with more information.

vamshin:
This is now available in Amazon Elasticsearch service from 7.9 onwards

PR #228, opened 2020-09-16
Add existsQuery method implementation to KNNVectorFieldType

jmazanec15:
*Issue #, if available:*
#197 

*Description of changes:*
Adds existQuery functionality support to KNNVectorFieldType.

This change allows a user to get the documents with a k-NN field like this:
```
{
    "query": {
       "exists": {
       "field": "imageVector"
     }
  }
}
```

Additionally, added tests to confirm functionality.

### DIFF
--- a/src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNESIT.java
+++ b/src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNESIT.java
@@ -29,6 +29,7 @@ import org.elasticsearch.index.query.ExistsQueryBuilder;
 import org.elasticsearch.rest.RestStatus;
 
 import java.io.IOException;
+import java.util.Arrays;
 import java.util.List;
 
 import static org.hamcrest.Matchers.containsString;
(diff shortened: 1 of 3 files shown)


src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNESIT.java
```diff
@@ -325,4 +329,36 @@ public void testInvalidQueryHnswAlgoParams() {
                 createKnnIndexMapping(FIELD_NAME, 2)));
         assertThat(ex.getMessage(), containsString("Failed to parse value [-1] for setting [index.knn.algo_param.ef_search]"));
     }
+
+    public void testExistsQuery() throws IOException {
```
vamshin, line 334: How about we declare 2 different knn fields(k1, k2) in the mappings with below scenario

Have 1 document with both the k1,k2.
Have 3 docs with only k1
Have 2 docs with only k2
Have 1 doc with neither k1 nor k2

With exists query for k1, assert hits=4
With exists query for k2, assert hits=3

jmazanec15: Makes sense. Updated the test cases.

vamshin (approved):
LGTM!

PR #233, opened 2020-09-21
Add stats for custom scoring feature

jmazanec15:
*Issue #, if available:*

*Description of changes:*
Adds a few statistics for custom scoring feature. The following statistics are added:
#### script_compilations
The number of times the knn script is compiled. This value should only be 0 or 1 most of the time. However, if the cache containing the compiled scripts is filled, it may cause the script to be recompiled.

#### script_compilation_errors
The number of errors during script compilation.

#### script_query_requests
The number of query requests that use the k-NN score script.

#### script_query_errors
The number of errors that have occurred during use of the k-NN score script.

Along with adding the stats, I updated the README and added another stats integration test to verify the changes.

I think in the future, it will probably make sense to refactor the stats api into different sections for custom scoring and approximate k Nearest Neighbor search. I will create an issue to do this.


src/test/java/com/amazon/opendistroforelasticsearch/knn/plugin/action/RestKNNStatsHandlerIT.java
```diff
@@ -164,6 +170,62 @@ public void testInvalidNodeIdStats() throws Exception {
         assertEquals(0, nodeStats.size());
     }
 
+    /**
+     *  Test checks that script stats are properly updated
+     */
+    public void testScriptStats() throws Exception {
```
vamshin: Can we add similar test with multiple shards?

README.md
```diff
@@ -352,6 +352,18 @@ The total time in nanoseconds it has taken to load items into cache (cumulative)
...
+#### script_compilations
+The number of times the knn script is compiled. This value should only be 0 or 1 most of the time. However, if the cache containing the compiled scripts is filled, it may cause the script to be recompiled.
+
+#### script_compilation_errors
+The number of errors during script compilation.
+
+#### script_query_requests
+The number of query requests that use the k-NN score script.
+
+#### script_query_errors
```
vamshin, line 364: I think these errors are counted at shard level. We could update the documentation accordingly. Also you might want to check if all the above metrics at shard level or index level

jmazanec15: Yes, they are at shard level. I updated README

src/test/java/com/amazon/opendistroforelasticsearch/knn/plugin/action/RestKNNStatsHandlerIT.java
jmazanec15: Sure

vamshin (approved):
LGTM! Thanks for the stats.

PR #235, opened 2020-09-22
Fix script statistics flaky test case

jmazanec15:
*Issue #, if available:*

*Description of changes:*
Script compilation count statistic in tests should be one. However, there are some cases where it may be 2. This change allows the test case to pass when compilation count is either 1 or 1 greater than what it was before a given query.


src/test/java/com/amazon/opendistroforelasticsearch/knn/plugin/action/RestKNNStatsHandlerIT.java
```diff
@@ -205,7 +207,8 @@ public void testScriptStats_singleShard() throws Exception {
                 StatNames.SCRIPT_QUERY_REQUESTS.getName())
         );
         nodeStats = parseNodeStatsResponse(EntityUtils.toString(response.getEntity()));
-        assertEquals(1, (int)(nodeStats.get(0).get(StatNames.SCRIPT_COMPILATIONS.getName())));
+        assertTrue((int)(nodeStats.get(0).get(StatNames.SCRIPT_COMPILATIONS.getName())) == 1 ||
+                (int)(nodeStats.get(0).get(StatNames.SCRIPT_COMPILATIONS.getName())) == initialScriptCompilations + 1);
```
vamshin: Why we need ```OR``` condition why not just 
```(int)(nodeStats.get(0).get(StatNames.SCRIPT_COMPILATIONS.getName())) == initialScriptCompilations + 1)```

Comment (jmazanec15) [reply src/test/java/com/amazon/opendistroforelasticsearch/knn/plugin/action/RestKNNStatsHandlerIT.java]: Yes, this was an attempted workaround. SCRIPT_COMPILATIONS will only be incremented if the script has yet to be compiled.

A better solution would be to clear the script cache before running the tests so that it is guaranteed that the script gets recompiled and SCRIPT_COMPILATIONS gets incremented by 1. I will update the PR.

File: src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java [hunk 2]
```diff
@@ -543,6 +543,16 @@ protected void clearCache() throws Exception {
         updateClusterSettings(KNNSettings.KNN_CACHE_ITEM_EXPIRY_TIME_MINUTES, null);
     }
 
+    /**
+     * Clear script cache
+     *
+     * Remove k-NN script from script cache so that it has to be recompiled
+     */
+    protected void clearScriptCache() throws Exception {
+        updateClusterSettings("script.context.score.cache_expire", "0");
```
vamshin, line 552: Is this line required?

jmazanec15: Yes, this is what causes cache to evict the script. Here is documentation: https://www.elastic.co/guide/en/elasticsearch/reference/7.x/modules-scripting-using.html#modules-scripting-using-caching

vamshin: Got it. Thanks

vamshin (approved):
LGTM! Thanks for the fix.

Issue #227, opened 2020-09-16
suggest is not a SuggestField


paintmeyellow:
Why when I am using `"index.knn = true"` I got error `suggest is not a SuggestField`?

Here is the **mapping** of suggest field:
```
"suggest": {
    "type": "completion",
    "analyzer": "simple",
    "preserve_separators": true,
    "preserve_position_increments": false,
    "max_input_length": 50
}
```
Here is the problem part of **query**:
```
"suggest": {
    "prefix_suggest": {
        "text": "white",
        "completion": {
            "field": "suggest"
        }
    }
}
```

vamshin:
Hi @Nikitos096,

Could you please provide the complete index mapping and the query to reproduce at our end?

paintmeyellow:
> Hi @Nikitos096,
> 
> Could you please provide the complete index mapping and the query to reproduce at our end?

Yep, here you go:
```
PUT music
{
  "settings": {
    "index": {
      "knn": true
    }
  }, 
  "mappings": {
    "properties": {
      "suggest": {
        "type": "completion"
      }
    }
  }
}

PUT music/_doc/1?refresh
{
  "suggest" : {
    "input": [ "Nevermind", "Nirvana" ],
    "weight" : 34
  }
}

POST music/_search?pretty
{
  "suggest": {
    "song-suggest": {
      "prefix": "nir",        
      "completion": {         
          "field": "suggest"  
      }
    }
  }
}
```
Source: https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-suggesters.html#completion-suggester

vamshin:
Looks like KNNCodec is missing this implementation. Will dig into this and update the thread soon.

vamshin:
We are considering this as a bug and we would prioritize the fix accordingly.

PR #236, opened 2020-09-24
Fix PostingsFormat in KNN Codec

vamshin:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/k-NN/issues/227

*Description of changes:*
Elasticsearch initiates PostingsFormat based on the MapperService. For KNNCodec, we do not pass MapperService and the PostingsFormat deviates from underlying format for the current Lucene version. Because of this some of the terms that use PostingsFormat could have incompatibility. For example  https://github.com/opendistro-for-elasticsearch/k-NN/issues/227.
Updated KNNCodec to rely on the right PostingsFormat used by underlying Codec of Elasticsearch.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN86Codec/KNN86Codec.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN86Codec/KNN86Codec.java
@@ -82,9 +82,14 @@ public final class KNN86Codec extends Codec {
      * approach of manually overriding.
      */
 
+    PostingsFormat postingsFormat = null;
+    public void setPostingsFormat(PostingsFormat postingsFormat) {
+        this.postingsFormat = postingsFormat;
+    }
+
     @Override
     public PostingsFormat postingsFormat() {
-        return getDelegatee().postingsFormat();
+        return this.postingsFormat;
     }
 
     @Override
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN86Codec/KNN86Codec.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN86Codec/KNN86Codec.java
@@ -89,6 +89,9 @@ public final class KNN86Codec extends Codec {
 
     @Override
     public PostingsFormat postingsFormat() {
+        if (this.postingsFormat == null) {
+            return getDelegatee().postingsFormat();
+        }
         return this.postingsFormat;
     }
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN86Codec/KNN86Codec.java
```diff
@@ -82,9 +82,17 @@ public DocValuesFormat docValuesFormat() {
      * approach of manually overriding.
      */
 
+    PostingsFormat postingsFormat = null;
```
jmazanec15: Why declare is here?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/KNNCodecService.java
```diff
@@ -45,4 +47,8 @@ public Codec codec(String name) {
         }
         return codec;
     }
+
+    public void setPostingsFormat(PostingsFormat postingsFormat) {
+        ((KNN86Codec)codec("")).setPostingsFormat(postingsFormat);
```
jmazanec15, line 52: Why is  this string empty?

jmazanec15 (commented):
Could we add test case confirming the fix works? Other than the suggests field, what other functionality does this bug break?

"Elasticsearch initiates PostingsFormat based on the MapperService. For KNNCodec, we do not pass MapperService"

Do you have code link for where Elasticsearch initiates PostingsFormat based on MapperService? Also, why can't we pass MapperService to KNNCodec?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/KNNCodecService.java
vamshin: ```codec``` function does not use this String.  We could pass any string here. Idea is to get the KNNCodec from this function

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN86Codec/KNN86Codec.java
vamshin: Fixed.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/KNNCodecService.java
jmazanec15: Makes sense

vamshin:
> Could we add test case confirming the fix works? Other than the suggests field, what other functionality does this bug break?
> 
> "Elasticsearch initiates PostingsFormat based on the MapperService. For KNNCodec, we do not pass MapperService"
> 
> Do you have code link for where Elasticsearch initiates PostingsFormat based on MapperService? Also, why can't we pass MapperService to KNNCodec?

1. Manually verified the fix.  I will add test case part of this PR https://github.com/opendistro-for-elasticsearch/k-NN/issues/245

2. Sure. This code is part of CodecService. Please refer this link https://github.com/elastic/elasticsearch/blob/v7.9.1/server/src/main/java/org/elasticsearch/index/codec/CodecService.java#L49-L56

jmazanec15 (approved):
LGTM thanks!

Issue #239, opened 2020-10-13
KNN graphs occasionally are indexed with wrong spaceType


jmazanec15:
Occasionally, an index's graphs that are supposed to be of one space type, gets indexed as a different type. This leads to poor recall.

The root cause appears to be in these lines:
1. https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java#L116
2. https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java#L121
3. https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java#L100

Here, we update the default field type after the default field type has been copied into a new field type for the builder class. In some cases, this leads to a cosine/l2 index's graphs being indexed with the space type that was set previously for `Defaults.FIELD_TYPE`

PR #240, opened 2020-10-13
Refactor KNNVectorFieldMapper

jmazanec15:
*Issue #, if available:*
#239 

*Description of changes:*
PR refactors KNNVectorFieldMapper in a variety of ways. 

First, it refactors KNNVectorFieldMapper to utilize ParametrizedFieldMapper. This way, we do not need to parse the field parameters manually. 

Second, it refactors KNNVectorFieldMapper to build a new fieldType in parseCreateField, as opposed to updating `Defaults.FIELD_TYPE` inside the builder logic. This will prevent race condition bugs seen in #239.

Third, it refactors the ordering of the logic inside of KNNVectorFieldMapper to follow the convention of other Elasticsearch field mappers.


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -344,6 +312,12 @@ public void parse(ParseContext context) throws IOException {
             array[i++] = f;
         }
 
+        FieldType fieldType = new FieldType(Defaults.FIELD_TYPE);
+        fieldType.putAttribute(KNNConstants.SPACE_TYPE, spaceType);
+        fieldType.putAttribute(KNNConstants.HNSW_ALGO_M, String.valueOf(m));
+        fieldType.putAttribute(KNNConstants.HNSW_ALGO_EF_CONSTRUCTION, String.valueOf(efConstruction));
```
vamshin: How about reading these algo params from the indexSettings here? Idea is to lazy load these settings to ensure the settings are available to read from IndexSettings.

jmazanec15: This would cause index settings to be read everytime a document is indexed. I refactored to build fieldType to the KNNVectorFieldMapper constructor and read the settings from BuilderContext during Builder.build()

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -122,59 +131,69 @@ public Builder(String name, String spaceType, int m, int efConstruction) {
...
         @Override
         public KNNVectorFieldMapper build(BuilderContext context) {
-            return new KNNVectorFieldMapper(name, new KNNVectorFieldType(buildFullName(context), meta.getValue(),
-                    dimension.getValue()), multiFieldsBuilder.build(this, context),
-                    ignoreMalformed(context), spaceType, m, efConstruction, copyTo.build(), this);
-        }
-    }
+            if (spaceType == null) {
+                spaceType = getSpaceType(context.indexSettings());
+            }
```
vamshin: Why are we storing state in Builder class? Why not get a fresh copy every time by reading from indexSettings?

jmazanec15: Makes sense. Updated.

vamshin (approved):
LGTM! Thanks for refactoring the KNNVectorFieldMapper and keeping it consistent with other Elasticsearch's Mapper. This will help in smooth migration for future releases.

Issue #255, opened 2020-10-20
Warmup API causes duplicate graphs to be loaded into memory


jmazanec15:
For background, the Plugin uses a cache to store references to the HNSW graphs used to perform k-NN search. The key for this cache is the path to the hnsw file.

On systems that use symlinks for the data path, [search](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java#L73) will load the graphs into the cache with the key being the absolute path to the data, not the symlink path.

However, [warmup](https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java#L108) will use the symlink path.

This leads the warmup api to load duplicate graphs into the cache.

The fix will be to make the paths the warmup api the same as the paths used in search.

PR #256, opened 2020-10-20
Change path warmup API uses to load graphs into cache

jmazanec15:
*Issue #, if available:*
#255 

*Description of changes:*
Update logic in warmup to get data directory in the same way that KNNWeight.scorer does.

Link to scorer code: https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java#L73


vamshin (approved):
LGTM!

Issue #264, opened 2020-11-02
Add binary Hamming distance to custom scoring


jmazanec15:
The Hamming distance measures the number of positions that are different between 2 equal length sequences of characters. This feature will look to support Hamming distance for binary sequences within the custom scoring logic.

A user should be able to ingest documents that have a term that can be interpreted as binary data and then apply the k-NN custom scoring script in order to get the k most similar results based on this distance metric. In order to implement this, we will need to determine which Elasticsearch types we will use to represent binary data. For instance, Elasticsearch has the [binary field type](https://www.elastic.co/guide/en/elasticsearch/reference/current/binary.html), that uses Base64 encoded values. We will also look into supporting integral types (long, int, short, byte) and their related arrays. These may offer performance improvements over Base64.

Additional enhancements to this feature include supporting fixed radius nearest neighbor search, which will return all documents that fall within a certain radius of the query data.

luyuncheng:
i see there is [space_bit_hamming](https://github.com/nmslib/nmslib/blob/master/similarity_search/include/factory/space/space_bit_hamming.h) in nmslib. is there any possible use this space in k-NN?

jmazanec15:
Hi @luyuncheng I think we could investigate adding that functionality. Because this issue is dedicated to adding Hamming distance in custom scoring, I am going to close it. Could you please open a separate issue for integrating nmslib's hamming bit space into the plugin?

PR #267, opened 2020-11-12
Support for hamming bit distance in custom scoring

jmazanec15:
*Issue #, if available:*
#264 

*Description of changes:*
This PR extends custom scoring to support hamming distance on binary data. The PR allows hamming distance to be applied to Long and Binary Elasticsearch fields. With the Long fieldType, a user will be able to encode up to 64 bits of data as a signed long using Elasticsearch's Long fieldtype. Here is an example query:
```
POST /myindex/_search
{
 "size": 4,
 "query": {
   "script_score": {
     "query": {
       "match_all": {}
     },
     "script": {
       "source": "knn_score",
       "lang": "knn",
       "params": {
         "field": "my_long",
         "query_value": 170,
         "space_type": "hammingbit"
       }
     }
   }
 }
}
```
I experimented with using long arrays to extend the number of bits a user can encode past 64 bits. However, because Elasticsearch doc values are not guaranteed to be in the order they were ingested in, we would have to use the "_source" field to accomplish this, which would add performance penalties.

In addition to supporting the Long type, a user can also run hamming distance k-NN search with Elasticsearch's binary field type on arbitrarily sized data. The binary field type uses Base64 encoding scheme. An example query would look like this:
```
POST /myindex2/_search
{
 "size": 2,
 "query": {
   "script_score": {
     "query": {
       "match_all": {}
     },
     "script": {
       "source": "knn_score",
       "lang": "knn",
       "params": {
         "field": "my_binary",
         "query_value": "q83vQUI=",
         "space_type": "hammingbit"
       }
     }
   }
 }
}
```

A third possibility would be for a user to provide a hex encoded string as the binary representation of their data. However, because it would be difficult to validate this format during indexing, it could prove difficult for usage. For that reason, I did not include it in this PR. However, if there is community interest in that, I could look into it further.

I changed several other components of the custom scoring implementation in order to support types other than KNNVectors. Additionally, I moved as much of the query validation/parsing out of the execute portion of the scripts as I could to improve performance.

Lastly, I changed the parameter for passing the query value from "vector" to "query_value" in order to generalize parsing. I was hoping to get some thoughts/opinions on this.

If a user wants to perform an r-NN search, where results are only returned if they meet a certain score, they can use Elasticsearch's [min_score](https://www.elastic.co/guide/en/elasticsearch/reference/7.x/search-search.html#search-api-min-score) parameter. 

For testing, I created a single node cluster based off of ODFE 1.11, and installed my changes onto it. I ran the testing script attached to the PR a few times to validate recall and performance ([hamming_test.txt](https://github.com/opendistro-for-elasticsearch/k-NN/files/5528814/hamming_test.txt)). Additionally, I added several unit and integration tests.


src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScript.java
```diff
@@ -0,0 +1,168 @@
...
+        this.queryValue = queryValue;
+        this.field = field;
+        this.scoringMethod = scoringMethod;
+    }
+
+    /**
+     * KNNScoreScript with Long type. The query value passed in as well as the DocValues being searched over are
+     * expected to be Longs.
+     */
+    public static class Longs extends KNNScoreScript<Long> {
```
VijayanB: may be LongType?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/util/KNNConstants.java
```diff
@@ -23,5 +23,6 @@
     public static final String HNSW_ALGO_INDEX_THREAD_QTY = "indexThreadQty";
     public static final String L2 = "l2";
     public static final String COSINESIMIL = "cosinesimil";
+    public static final String BIT_HAMMING = "bit_hamming";
```
VijayanB: Q: cosinesimil is one word without underscore while bit_hamming is not, so any reason why it is different from cosinesimil?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScript.java
```diff
@@ -0,0 +1,168 @@
...
+                    BiFunction<Long, Long, Float> scoringMethod, SearchLookup lookup, LeafReaderContext leafContext) {
+            super(params, queryValue, field, scoringMethod, lookup, leafContext);
+        }
+    }
+
+    /**
+     * KNNScoreScript with BitSet type. The query value passed in as well as the DocValues being searched over
+     * are expected to be BitSets.
+     */
+    public static class BitSets extends KNNScoreScript<BitSet> {
```
VijayanB: same as above

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpaceFactory.java
```diff
@@ -0,0 +1,39 @@
...
+import com.amazon.opendistroforelasticsearch.knn.index.util.KNNConstants;
+import com.amazon.opendistroforelasticsearch.knn.plugin.stats.KNNCounter;
+import org.elasticsearch.index.mapper.MappedFieldType;
+
+/**
+ * Factory to create correct KNNScoringSpace based on the spaceType passed in.
+ */
+public class KNNScoringSpaceFactory {
+    public static KNNScoringSpace getSpace(String spaceType, Object query, MappedFieldType mappedFieldType) {
+        if (spaceType.equalsIgnoreCase(KNNConstants.BIT_HAMMING)) {
```
VijayanB: It is always better to  do constant.equalsIgnoreCase(variable) to avoid null pointer exception.

VijayanB: Also you don't need else if, since you are returning inside if.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -73,6 +89,30 @@ public static float cosinesimil(float[] queryVector, float[] inputVector) {
...
+    }
+
+    /**
+     * This method calculates hamming distance on 2 longs
+     *
+     * @param queryLong query Long
+     * @param inputLong input Long
+     * @return hamming distance
+     */
+    public static float bitHamming(Long queryLong, Long inputLong) {
```
VijayanB: calculateHammingDistance?

src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java
```diff
@@ -356,6 +356,28 @@ protected void addDocWithNumericField(String index, String docId, String fieldNa
...
+
+        XContentBuilder builder = XContentFactory.jsonBuilder().startObject()
+                .field(fieldName, base64String)
+                .endObject();
+
+        request.setJsonEntity(Strings.toString(builder));
+
+        Response response = client().performRequest(request);
+
+        assertEquals(request.getEndpoint() + ": failed", RestStatus.CREATED,
```
VijayanB, line 377: do we need assert here? method says Add a single numeric field Doc to an index. If it is part of test method, may be we can assert inside actual test method instead of helper method. What do you think?

src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java
```diff
@@ -572,4 +594,26 @@ protected Request constructKNNScriptQueryRequest(String indexName, QueryBuilder
...
     }
+
+    protected Request constructKNNScriptQueryRequest(String indexName, QueryBuilder qb, Map<String, Object> params,
+                                                     int size) throws Exception {
+        Script script = new Script(Script.DEFAULT_SCRIPT_TYPE, KNNScoringScriptEngine.NAME, KNNScoringScriptEngine.SCRIPT_SOURCE, params);
+        ScriptScoreQueryBuilder sc = new ScriptScoreQueryBuilder(qb, script);
+        XContentBuilder builder = XContentFactory.jsonBuilder().startObject()
+                .field("size", size)
+                .startObject("query");
+        builder.startObject("script_score");
```
VijayanB, line 605: is it possible to refer from constant?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
```diff
@@ -0,0 +1,209 @@
...
+                throw new IllegalArgumentException("Incompatible field_type for hamming space. The field type must " +
+                        "of type long or binary.");
+            }
+        }
+
+        @Override
+        @SuppressWarnings("unchecked")
+        public ScoreScript getScoreScript(Map<String, Object> params, String field, SearchLookup lookup,
+                                          LeafReaderContext ctx) throws IOException {
+            if (isLongFieldType(fieldType)) {
```
VijayanB: no need of else if since you are returning inside if.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
```diff
@@ -0,0 +1,209 @@
...
+     * @param field Fieldname
+     * @param lookup SearchLookup
+     * @param ctx ctx LeafReaderContext to be used for scoring documents
+     * @return ScoreScript for this query
+     * @throws IOException throws IOException if ScoreScript cannot be constructed
+     */
+    public abstract ScoreScript getScoreScript(Map<String, Object> params, String field, SearchLookup lookup,
+                                               LeafReaderContext ctx) throws IOException;
+
+    public static class L2 extends KNNScoringSpace {
```
VijayanB: I believe L2 is Euclidean distance, may be name class as Euclidean?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
```diff
@@ -0,0 +1,209 @@
...
+     * Prepare the query and the scoring method for the given FieldType and similarity space. When preparing these
+     * properties, the fieldType and the spaceType will dictate if they are validated and how they are processed. This
+     * method has the responsibility of ensuring that the query and scoring method that will be used when scoring the
+     * docs in an index is compatible.
+     *
+     * @param query Raw query object passed in to be validated and processed for the given similarity space
+     */
+    protected abstract void prepareQuery(Object query);
+
+    protected boolean isLongFieldType(MappedFieldType fieldType) {
```
VijayanB: why isLongFieldType/isBinaryFieldType/isKNNVectorFieldType are protected?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
```diff
@@ -0,0 +1,209 @@
...
+
+import static org.elasticsearch.index.mapper.NumberFieldMapper.NumberType.LONG;
+
+/**
+ * KNNScoringSpace is used to validate/prepare a user provided query and similarity space for knn scripting execution
+ * and provide the correct KNNScoringScript.
+ */
+public abstract class KNNScoringSpace {
+
+    protected Object processedQuery;
```
VijayanB: why not make processedQuery as method and do validation inside it? it is kind of dangerous to call protected method inside constructor because you are calling an instance method before creating an instance.
IMO, KNNScoringSpace can be made interface and make helper method in a different class.

VijayanB (commented):
I haven't finished review yet, thought i will post so that some of the questions can be answered .

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpaceFactory.java
```diff
@@ -0,0 +1,39 @@
...
+
+import com.amazon.opendistroforelasticsearch.knn.index.util.KNNConstants;
+import com.amazon.opendistroforelasticsearch.knn.plugin.stats.KNNCounter;
+import org.elasticsearch.index.mapper.MappedFieldType;
+
+/**
+ * Factory to create correct KNNScoringSpace based on the spaceType passed in.
+ */
+public class KNNScoringSpaceFactory {
+    public static KNNScoringSpace getSpace(String spaceType, Object query, MappedFieldType mappedFieldType) {
```
VijayanB: may be createScoringSpace or create instead of get?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
```diff
@@ -0,0 +1,209 @@
...
+        @SuppressWarnings("unchecked")
+        public ScoreScript getScoreScript(Map<String, Object> params, String field, SearchLookup lookup,
+                                          LeafReaderContext ctx) throws IOException {
+            if (isLongFieldType(fieldType)) {
+                return new KNNScoreScript.Longs(params, (Long) this.processedQuery, field,
+                        (BiFunction<Long, Long, Float>) this.scoringMethod, lookup, ctx);
+            } else if (isBinaryFieldType(fieldType)) {
+                return new KNNScoreScript.BitSets(params, (BitSet) this.processedQuery, field,
+                        (BiFunction<BitSet, BitSet, Float>) this.scoringMethod, lookup, ctx);
+            } else {
```
VijayanB: is this even required since the fieldType is already tested in parseQuery?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/util/KNNConstants.java
jmazanec15: Good catch. I will change to bithamming.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScript.java
jmazanec15: Yes, I think that makes sense. Will update.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScript.java
jmazanec15: ack

src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java
jmazanec15: I left the assert in this function because the test methods are not actually testing this functionality. The test methods are expecting this functionality to work, so I included it in here. Do you think it would be better practice to return the response and check it in the tests?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
jmazanec15: Makes sense, will update

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpaceFactory.java
jmazanec15: Makes sense

jmazanec15: Good catch. Will fix

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpaceFactory.java
jmazanec15: Makes sense. Will update.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
jmazanec15: Makes sense. Will update

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
jmazanec15: I think we should keep as L2 because thats how we refer to it in user facing APIs. Additionally, nmslib refers to it as L2, so I think we should maintain consistency.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
jmazanec15: My intention was to allow subclasses to utilize these methods, while preventing them from being used outside of the class. Do you think they should have a different access modifier?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
jmazanec15: I don't believe so.

src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java
jmazanec15: This is a part of Elasticsearch Query DSL. I couldnt find a constant I could refer to for this.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
jmazanec15: Agree with first point.

For second point, could you explain more why you think that?

src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java
VijayanB: I am not sure what is the convention, in my opinion, it will be clear to differentiate testing vs helping method for maintenance.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
VijayanB: got it. i was concerned about having a class name with two characters. If it makes sense, i totally agree. Also, what do you think about L2ScoringSpace / CosineSimilarityScoringSpace ?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
VijayanB: Generally subclass are used when it naturally depicts parent -> child relationship where there are logics which are common to be inherited and some to be overridden.  Here, there is no strategy/algorithm that is common while child class can behave differently. I only see some helper methods that will be used by child class and the common method(getScoreScript) which will be called by outside is abstract.  Also, any reason why made L2/Hamming as inner class ? Interface is preferred if intention is that L2ScoringSpace/BitHammingScoringSpace are all ScoringSpace.
Did i convince you?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScript.java
```diff
@@ -0,0 +1,168 @@
...
+        @Override
+        public void setDocument(int docId) {
+            try {
+                this.vectorExist = this.binaryDocValuesReader.advanceExact(docId);
+            } catch (IOException e) {
+                throw new UncheckedIOException(e);
+            }
+        }
+
+        public KNNVectorType(Map<String, Object> params, float[] queryValue, String field,
```
VijayanB: nit: can we move this close to class?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScriptFactory.java
```diff
@@ -0,0 +1,88 @@
...
+
+public class KNNScoreScriptFactory implements ScoreScript.LeafFactory {
+    private final Map<String, Object> params;
+    private final SearchLookup lookup;
+    private String similaritySpace;
+    private String field;
+    private Object query;
+    private KNNScoringSpace knnScoringSpace;
+
+    public KNNScoreScriptFactory(Map<String, Object> params, SearchLookup lookup) {
```
VijayanB, line 34: What do you think about following format?    
  
        public KNNScoreScriptFactory(Map<String, Object> params, SearchLookup lookup) {
         KNNCounter.SCRIPT_QUERY_REQUESTS.increment();
         this.params = params;
         this.lookup = lookup;
         this.field = getValue(params, "field")
         similaritySpace = getValue(params, "space_type")
         this.knnScoringSpace = KNNScoringSpaceFactory.create(similaritySpace, this.query,
                lookup.doc().mapperService().fieldType(this.field));

    }

     private Object getValue(Map<String, Object> params, String fieldName) {
        final Object value = params.get(fieldName);
        if (value != null) 
         return value.toString();
       KNNCounter.SCRIPT_QUERY_ERRORS.increment();
       throw new IllegalArgumentException("Missing parameter ["+ fieldName +"]");
      }

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScriptFactory.java
```diff
@@ -0,0 +1,88 @@
...
+
+        final Object queryValue = params.get("query_value");
+        if (queryValue == null) {
+            KNNCounter.SCRIPT_QUERY_ERRORS.increment();
+            throw new IllegalArgumentException("Missing parameter [query_value]");
+        }
+        this.query = queryValue;
+    }
+
+    public boolean needs_score() {
```
VijayanB, line 56: needsScore

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScript.java
```diff
@@ -0,0 +1,168 @@
...
+         * This function calculates the similarity score for each doc in the segment.
+         *
+         * @param explanationHolder A helper to take in an explanation from a script and turn
+         *                          it into an {@link org.apache.lucene.search.Explanation}
+         * @return score for the provided space between the doc and the query
+         */
+        @Override
+        public double execute(ScoreScript.ExplanationHolder explanationHolder) {
+            ScriptDocValues.BytesRefs scriptDocValues = (ScriptDocValues.BytesRefs) getDoc().get(this.field);
+            if (scriptDocValues.size() == 0) {
```
VijayanB: nit: does scriptDocValues has isEmpty method?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScriptFactory.java
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * For each segment, supply the KNNScoreScript that should be run on the values returned from the fetch phase.
+     * Because the method to score the documents was set during Factory construction, the scripts are agnostic of
+     * the similarity space. The KNNScoringSpace will return the correct script, given the query, the field type, and
+     * the similarity space.
+     *
+     * @param ctx LeafReaderContext for the segment
+     * @return ScoreScript to be executed
+     */
+    @Override // called number of segments times
```
VijayanB: nit: is this required? if so, can we move to block comment?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScript.java
jmazanec15: Yes it does. I will update.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScript.java
jmazanec15: Yes we can

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScriptFactory.java
jmazanec15: Thats actually a method in the LeafFactory interface. So, we can't change.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScriptFactory.java
jmazanec15: Good point, will update.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScriptFactory.java
jmazanec15: Yeah, I think overall that is cleaner. I will update to follow that, with a few minor tweaks.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
jmazanec15: I decided not to add ScoringSpace to the end to remove redundancy. For instance, when using it in the ScoringSpaceFactory, the call looks like `KNNScoringSpace.HammingBit`. I thought this was more concise and cleaner than doing `KNNScoringSpace.HammingBitScoringSpace`. What are your thoughts?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringSpace.java
jmazanec15: Yes, I think you did. I will update to use an interface.

src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java
jmazanec15: I think some of our other methods in this test case follow this convention. I think, to stay on track for this PR, we follow this convention. And, if we decide that we need to refactor testing logic, we take up another PR for it.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScriptFactory.java
VijayanB: i see. Can you add @override annotation ?

VijayanB:
Can you update the description to use  "space_type": "hammingbit"?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScriptFactory.java
jmazanec15: ack

jmazanec15:
> Can you update the description to use "space_type": "hammingbit"?

Done

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -1,9 +1,24 @@
...
+
 package com.amazon.opendistroforelasticsearch.knn.plugin.script;
 
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
-import java.util.ArrayList;
+import java.math.BigInteger;
 
 public class KNNScoringUtil {
```
vamshin, line 23: Should we merge this class with KNNScoringSpaceUtil?

src/test/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScriptScoringIT.java
```diff
@@ -308,4 +322,169 @@ public void testKNNScoreforNonVectorDocument() throws Exception {
         assertEquals(0.33333, scores.get(0), 0.001);
         assertEquals(Float.MIN_VALUE, scores.get(1), 0.001);
     }
+
+    @SuppressWarnings("unchecked")
+    public void testHammingScriptScore_Long() throws Exception {
```
vamshin, line 327: How about we add a testcase in such a way a segment contains docs with field name and with out the field name.   This would test how the scoring is handled for the segments with the above combination.  

We could try forcemerge and then assert that docs with out the similarity field are not in the top hits. One test for long and one test for base64 should be good?

vamshin (commented):
Overall looks good to me. Couple of comments

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
jmazanec15: So, I thought it might be worth keeping KNNScoringSpaceUtil and KNNScoringUtil separate. KNNScoringSpaceUtil contains utility functions for parsing/preparing queries, KNNScoringUtil contains the methods that compute the distance for a particular space.

I could rename KNNScoringUtil to KNNDistanceUtil if that sounds better? What do you think @vamshin @VijayanB ?

src/test/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScriptScoringIT.java
jmazanec15: Sure, so at the moment, I add 4 docs with the field and then 3 docs without the field, and then confirm the docs without the field arent in the results. What I will do is put in a refresh in between the addition of the docs with the field and the docs without the field, to confirm that a segment either has only docs with the field or only docs without the field. Then, IO will confirm the docs without the field arent in the results. Then I will forcemerge to 1 segment, and repeat the experiment.

vamshin (approved):

VijayanB (approved):
Thanks for making changes. LGTM.

Issue #270, opened 2020-11-25
plugin support for Elasticsearch version 7.10.0(opendistro 1.12)


vamshin:
plugin support for Elasticsearch version 7.10.0(opendistro 1.12)

PR #271, opened 2020-11-25
k-NN plugin support for Elasticsearch version 7.10.0

vamshin:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/k-NN/issues/270

*Description of changes:*
k-NN plugin support for Elasticsearch version 7.10.0


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -78,9 +79,9 @@ private static KNNVectorFieldMapper toType(FieldMapper in) {
                 m -> toType(m).stored, false);
         private final Parameter<Boolean> hasDocValues = Parameter.boolParam("doc_values", false,
                 m -> toType(m).hasDocValues,  true);
-        private final Parameter<Integer> dimension = new Parameter<>(KNNConstants.DIMENSION, false, -1,
-                (n, o) -> {
-                            int value = XContentMapValues.nodeIntegerValue(o);
+        private final Parameter<Integer> dimension = new Parameter<>(KNNConstants.DIMENSION, false, () -> -1,
+                (n, c, o) -> {
+                            int value = (o == null ? null : XContentMapValues.nodeIntegerValue(o));
```
VijayanB: if o is null, then i don't think we can assign null to value since it is primitive.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN87Codec/KNN87Codec.java
```diff
@@ -0,0 +1,138 @@
+/*
+ *   Copyright 2019 Amazon.com, Inc. or its affiliates. All Rights Reserved.
```
VijayanB: Copyright 2020

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
vamshin: Good catch. Fixed.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN87Codec/KNN87Codec.java
vamshin: Fixed. Thanks

gradle.properties
```diff
@@ -13,4 +13,4 @@
 #   permissions and limitations under the License.
 #
 
-version=1.6
+version=1.11.0
```
jmazanec15, line 16: What is this for?

jmazanec15 (commented):
Just to confirm, this PR is not meant to upgrade ODFE to 1.12, correct? That will be done in a different PR?

gradle.properties
vamshin: Actually this variable is not in use. It was pointing to older version. So i updated to latest version. This variable could be used in build.gradle file.

vamshin:
> Just to confirm, this PR is not meant to upgrade ODFE to 1.12, correct? That will be done in a different PR?


right. This was to upgrade to 7.10.

gradle.properties
jmazanec15: I see makes sense.

jmazanec15 (approved):

VijayanB (approved):
LGTM

Issue #279, opened 2020-12-09
Testing workflow fails due to CMake Error


jmazanec15:
Failure: https://github.com/opendistro-for-elasticsearch/k-NN/pull/278/checks?check_run_id=1526708387

Error:
```
 CMake Error at CMakeLists.txt:34 (find_path):
  Could not find NMS_REPO_DIR using the following files: similarity_search
```
For this action, we use ubuntu-latest, which corresponds to ubuntu 18. Looks like default CMake version changes from 3.17 to 3.19 in this commit: https://github.com/actions/virtual-environments/commit/5987ab0f60c8788d73e883406c9a17e7ea054194

Will need to update this line in CMakeLists.txt: https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/jni/CMakeLists.txt#L34

PR #280, opened 2020-12-09
Fix find_path bug in CMakeLists

jmazanec15:
*Issue #, if available:*
#279 

*Description of changes:*
This PR fixes a build failure when using CMake >= 3.18. During the build process, we use `find_path` to check if the nmslib submodule has already been pulled. If it has not, we pull it (in line 37). Recently, Ubuntu 18 in Github actions upgraded from Cmake 3.17 to 3.19. 

For some reason, I added the REQUIRED parameter in find_path. This had no effect in Cmake < 3.18. In Cmake >= 3.18, this causes CMake process to fail: https://cmake.org/cmake/help/v3.18/command/find_path.html. So, the fix is to remove REQUIRED parameter.

I verified the change works for CMake 3.19 on Ubuntu 18. I also verified that the fix does not break functionality for 3.17 on my Mac. 

### DIFF
--- a/jni/CMakeLists.txt
+++ b/jni/CMakeLists.txt
@@ -31,7 +31,7 @@ set(KNN_INDEX KNNIndexV2_0_6)
 set(KNN_PACKAGE_NAME opendistro-knnlib)
 
 # Check if similarity search exists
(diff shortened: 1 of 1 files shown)


vamshin (approved):
LGTM!

VijayanB (approved):

Issue #213, opened 2020-09-04
Expose similarity score functions as part of custom scoring


vamshin:
Expose functions that evaluate the similarity score like Euclidean, cosine etc so that users can perform their own scoring.

vamshin:
https://discuss.opendistrocommunity.dev/t/combining-knn-score-with-keyword-query/3600/3

ngadkari:
This will be a pretty useful feature for mixing knn and keyword search results. Is there an ETA for this feature

timforr:
Any updates on if this will make it in to the next release?

vamshin:
Hi @timforr, @VijayanB is working on the PR and we should be ready for odfe1.13 release. 

As a side note we already have custom scoring exposed from odfe 1.11  https://opendistro.github.io/for-elasticsearch-docs/docs/knn/#custom-scoring

Issue #222, opened 2020-09-15
Add README section for Custom scoring feature


vamshin:
Add README section for Custom scoring feature 
Parent PR:- https://github.com/opendistro-for-elasticsearch/k-NN/pull/196

jmazanec15:
Related PR: https://github.com/opendistro/for-elasticsearch-docs/pull/396

In order to standardize docs, we should host all usage information in the documentation located here: https://opendistro.github.io/for-elasticsearch-docs/docs/knn/

PR #281, opened 2020-12-11
Support k-NN similarity functions in painless scripting

VijayanB:
*Issue #213

Implement IndexFieldDataBuilder to allow painless scripting to indenitfy KNNScriptDocValues
Add dependency to PainlessExtension to add custom scoring methods to be exposed for usage by users inside painless scripting.


VijayanB:
Following methods are whitelisted:
l2Squared
cosineSimilarity
cosineSimilarityOptimized

@vamshin @jmazanec15  we can discuss here if we want to rename the method name for better reach

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -207,6 +210,12 @@ public Query termQuery(Object value, QueryShardContext context) {
         public int getDimension() {
             return dimension;
         }
+
+        @Override
+        public IndexFieldData.Builder fielddataBuilder(String fullyQualifiedIndexName, Supplier<SearchLookup> searchLookup) {
```
jmazanec15, line 215: fielddataBuilder should be fieldDataBuilder based on IndexFieldData

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -50,7 +90,7 @@ public static float l2Squared(float[] queryVector, float[] inputVector) {
      * @return cosine score
      */
     public static float cosinesimilOptimized(float[] queryVector, float[] inputVector, float normQueryVector) {
-        float dotProduct = 0.0f;
+            float dotProduct = 0.0f;
```
jmazanec15: Why is this indented?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -64,6 +104,35 @@ public static float cosinesimilOptimized(float[] queryVector, float[] inputVecto
         return (float) (dotProduct / (Math.sqrt(normalizedProduct)));
     }
 
+    /**
+     * Whitelisted method for users that can be used script to avoid repeated calculation of normalization
+     * for query vector for each filtered documents
+     * example
+     *  "script": {
+     *         "source": "cosineSimilarity(params.query_vector, docs[field], 1.0) ",
```
jmazanec15, line 126: This should be cosineSimilarityOptimized, no?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -64,6 +104,35 @@ public static float cosinesimilOptimized(float[] queryVector, float[] inputVecto
...
+     *           "field": "my_dense_vector"
+     *         }
+     *       }
+     *
+     * @param queryVector query vector
+     * @param docValues script doc values
+     * @param normQueryVector normalized query vector value.
+     * @return cosine score
+     */
+    public static float cosineSimilarityOptimized(
```
jmazanec15: I think this interface is vague: cosineSimilarityOptimized does not give a user any information about how this function differs from cosineSimilarity. If cosineSimilarity can be optimized, why even provide cosineSimilarity?

The Cosine similarity is A . B / (||A|| x ||B||). The difference between the two functions is that cosineSimilarityOptimized has the user pass in both the query vector and the magnitude of the query vector. Assuming the query vector is A, ||A|| does not change throughout the query. So time is saved by computing ||A|| separately and then passing it into the function. Also, I think normQueryVector may not be the appropriate term. Normalization refers to the process of making a vector have a magnitude of one.

My proposal is to switch it to the following interface:
```
cosineSimilarity(List <Number> queryVector, KNNVectorScriptDocValues docValues, Number queryVectorMagnitude)
```

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -64,6 +104,35 @@ public static float cosinesimilOptimized(float[] queryVector, float[] inputVecto
         return (float) (dotProduct / (Math.sqrt(normalizedProduct)));
     }
 
+    /**
+     * Whitelisted method for users that can be used script to avoid repeated calculation of normalization
```
jmazanec15: Nit: "Whitelisted cosineSimilarity method that can be used in a script to avoid repeated calculation of normalization for the query vector.

Example:
...
"

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
```diff
@@ -0,0 +1,69 @@
...
+        }
+        docExists=false;
+    }
+
+    public float[] getValue() throws IOException{
+        if (!docExists) {
+            throw new IllegalArgumentException("no value found for the corresponding doc ID");
+        }
+        BytesRef value = binaryDocValues.binaryValue();
+        Objects.requireNonNull(value);
```
jmazanec15: Is this redundant with line 46?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
```diff
@@ -0,0 +1,69 @@
...
+        ByteArrayInputStream byteStream = new ByteArrayInputStream(value.bytes, value.offset, value.length);
+        ObjectInputStream objectStream = new ObjectInputStream(byteStream);
+        try {
+            return (float[]) objectStream.readObject();
+        }catch (ClassNotFoundException e) {
```
jmazanec15: space before catch

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
```diff
@@ -0,0 +1,69 @@
...
+    private boolean docExists;
+
+    public KNNVectorScriptDocValues(BinaryDocValues binaryDocValues) {
+        this.binaryDocValues = binaryDocValues;
+    }
+
+    @Override
+    public void setNextDocId(int docId) throws IOException {
+        if (binaryDocValues.advanceExact(docId)) {
+            docExists = true;
```
jmazanec15, line 41: Is this thread safe? I am wondering if this is the correct place to determine if a value exists for that doc.

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorDVLeafFieldDataTests.java
```diff
@@ -0,0 +1,91 @@
...
+        KNNVectorDVLeafFieldData leafFieldData = new KNNVectorDVLeafFieldData(leafReaderContext.reader(), "");
+        assertEquals(0, leafFieldData.ramBytesUsed());
+    }
+
+    public void testGetBytesValues() {
+        KNNVectorDVLeafFieldData leafFieldData = new KNNVectorDVLeafFieldData(leafReaderContext.reader(), "");
+        expectThrows(UnsupportedOperationException.class,
+                () -> leafFieldData.getBytesValues());
+    }
+}
```
jmazanec15: nit: add endline

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorIndexFieldDataTests.java
```diff
@@ -0,0 +1,93 @@
...
+        expectThrows(UnsupportedOperationException.class,
+                () -> indexFieldData.sortField(null, null, null, false));
+    }
+
+    public void testNewBucketedSort() {
+
+        expectThrows(UnsupportedOperationException.class,
+                () -> indexFieldData.newBucketedSort(null, null, null, null, null, null, 0, null));
+    }
+}
```
jmazanec15: nit: add endline

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValuesTests.java
```diff
@@ -0,0 +1,85 @@
...
+        Assert.assertEquals(0, scriptDocValues.size());
+        scriptDocValues.setNextDocId(0);
+        Assert.assertEquals(1, scriptDocValues.size());
+    }
+
+    public void testGet() throws IOException {
+        expectThrows(UnsupportedOperationException.class,
+                () -> scriptDocValues.get(0));
+    }
+}
```
jmazanec15: add endline

jmazanec15 (commented):
A few comments. Still need to dig into the test cases more. One note, once #222 is resolved, we will need to add a section on this.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
VijayanB: Did you mean method name? if so, it is from parent class. i am overriding it here

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
VijayanB: ACk. Will apply reformat code in next commit.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
VijayanB: Ack.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
jmazanec15: Oh haha yes you are correct, my mistake

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
VijayanB: Ack

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
VijayanB: Ack

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
VijayanB: Ack. Applying reformat code.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
VijayanB: other scriptdocs like BInary, doesn't make this thread safe. Nevertheless, synchronized the methods to make sure we will not have race conditions.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
VijayanB: ACk

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValuesTests.java
VijayanB: Ack

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorIndexFieldDataTests.java
VijayanB: Ack

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorDVLeafFieldDataTests.java
VijayanB: ack

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
VijayanB: addressed your rename comment, hence keeping cosineSimilarity as overloaded.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
```diff
@@ -0,0 +1,72 @@
...
+                docExists = true;
+                return;
+            }
+            docExists = false;
+        }
+    }
+
+    public synchronized float[] getValue() throws IOException {
+        if (!docExists) {
+            throw new IllegalArgumentException("no value found for the corresponding doc ID");
```
jmazanec15: Why throw IllegalArgumentException? It seems like it maybe should be an IllegalStateException.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
```diff
@@ -0,0 +1,72 @@
...
+import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.util.BytesRef;
+import org.elasticsearch.index.fielddata.ScriptDocValues;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.io.ObjectInputStream;
+
+// This class is thread safe, since docExists is synchronized at an instance level
+public final class KNNVectorScriptDocValues extends ScriptDocValues<float[]> {
```
jmazanec15, line 27: Should this extend BinaryScriptDocValues<float[]>? Here is the [code](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/index/fielddata/ScriptDocValues.java#L440).

I think if we go this route, we do not have to have docExists boolean.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorDVLeafFieldData.java
```diff
@@ -0,0 +1,49 @@
...
+
+import org.apache.lucene.index.BinaryDocValues;
+import org.apache.lucene.index.LeafReader;
+import org.elasticsearch.index.fielddata.LeafFieldData;
+import org.elasticsearch.index.fielddata.ScriptDocValues;
+import org.elasticsearch.index.fielddata.SortedBinaryDocValues;
+
+import java.io.IOException;
+
+public class KNNVectorDVLeafFieldData implements LeafFieldData {
```
jmazanec15, line 12: Should this extend [AbstractBinaryDVLeafFieldData](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/index/fielddata/plain/AbstractBinaryDVLeafFieldData.java)?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -34,18 +37,55 @@
...
+     *
+     * @param queryVector query vector
+     * @param docValues   script doc values
+     * @return L2 score
+     */
+    public static float l2Squared(List<Number> queryVector, KNNVectorScriptDocValues docValues) {
+        float[] knnDocVector;
+        try {
+            knnDocVector = docValues.getValue();
+        } catch (Exception e) {
```
jmazanec15: Should we catch the specific exception here?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -34,18 +37,55 @@
...
+        return l2Squared(toFloat(queryVector), knnDocVector);
+    }
+
     /**
      * This method can be used script to avoid repeated calculation of normalization
      * for query vector for each filtered documents
      *
-     * @param queryVector query vector
-     * @param inputVector input vector
+     * @param queryVector     query vector
```
jmazanec15, line 100: Out of curiosity, any reason for these tabs here?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
jmazanec15: Makes sense

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -64,6 +104,35 @@ public static float cosinesimilOptimized(float[] queryVector, float[] inputVecto
...
+     *         "source": "cosineSimilarity(params.query_vector, docs[field], 1.0) ",
+     *         "params": {
+     *           "query_vector": [1, 2, 3.4],
+     *           "field": "my_dense_vector"
+     *         }
+     *       }
+     *
+     * @param queryVector     query vector
+     * @param docValues       script doc values
+     * @param normQueryVector normalized query vector value.
```
jmazanec15: The other part of my comment suggested changing this name to queryVectorMagnitude. normQueryVector makes it seem like the user should pass in the normalized query vector, not the magnitude of the query vector.

src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java
```diff
@@ -418,15 +419,15 @@ protected void deleteKnnDoc(String index, String docId) throws IOException {
      */
     protected void updateClusterSettings(String settingKey, Object value) throws Exception {
         XContentBuilder builder = XContentFactory.jsonBuilder()
-                       .startObject()
-                       .startObject("persistent")
-                       .field(settingKey, value)
-                       .endObject()
-                       .endObject();
+                .startObject()
```
jmazanec15, line 422: Which style tool did you use for formatting? I think it might make sense to add some kind of styling framework to the code so that we stay consistent (I am not suggesting adding it in this PR, but it might make sense to add in the future).

jmazanec15 (commented):
I think test cases look pretty good. I added a few additional comments in implementation.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
VijayanB: Thats a good point. That was my first attempt, but BinaryScriptDocValues is not public, hence we can't extend it.

VijayanB: Thats a good point. That was my first attempt, but BinaryScriptDocValues is not public, hence we can't extend it.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorDVLeafFieldData.java
VijayanB: same as above. The class is not public.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
VijayanB: Screen Shot 2020-12-16 at 10 27 38 AM

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
VijayanB: I see. I will update it, on side note, i don't think user will see this variable name since they will only see datatype.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
VijayanB: Ack

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
VijayanB: getValue can throw RuntimeException & IOException, i see three option here
1) Let docValues throw exception as it is and catch all individual exception here and return Float.Min like now,
2) Update getValue to throw only RuntimeException, and catch only RuntimeException here
3) Just update catch (Exception e) to catch ( RuntimeException | IOException e )

i don't see any difference in all three since at the end outcome is same, but if you have strong opinion on one vs other, i can change it . if you see any other option i can make change it as well.

src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java
VijayanB: i am using the one which comes with Intellij

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
jmazanec15: Ah, I see. That makes sense. In terms of docExists, I think we should follow the convention of using [count](https://github.com/elastic/elasticsearch/blob/master/server/src/main/java/org/elasticsearch/index/fielddata/ScriptDocValues.java#L493) instead. What do you think?

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
jmazanec15: Oh okay, makes sense. Yes, in the future I think we should use some kind of linter to keep style consistent.

src/test/java/com/amazon/opendistroforelasticsearch/knn/KNNRestTestCase.java
jmazanec15: Refer to above

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
jmazanec15: I see. Refer to comment below about count

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
VijayanB: We can do that too, but i kept consistent with our other implementation as here
https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoreScript.java#L138

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
VijayanB: i agree.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
jmazanec15: You are right. We probably don't need to make that change now.

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
jmazanec15: Makes sense, Option 1 us good with me

jmazanec15 (approved):
All looks good to me!

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
```diff
@@ -0,0 +1,72 @@
...
+    private final BinaryDocValues binaryDocValues;
+    private boolean docExists;
+
+    public KNNVectorScriptDocValues(BinaryDocValues binaryDocValues) {
+        this.binaryDocValues = binaryDocValues;
+    }
+
+    @Override
+    public void setNextDocId(int docId) throws IOException {
+        synchronized (this) {
```
vamshin: do we need ```synchronized``` here?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
```diff
@@ -0,0 +1,72 @@
...
+        try {
+            return (float[]) objectStream.readObject();
+        } catch (ClassNotFoundException e) {
+            throw new RuntimeException((e));
+        }
+    }
+
+    @Override
+    public int size() {
+        synchronized (this) {
```
vamshin: Do we need ```synchronized``` here?  Avoid synchronization in the places not needed as it could hamper performance.

src/test/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/PainlessScriptScoringIT.java
```diff
@@ -0,0 +1,124 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+public class PainlessScriptScoringIT extends KNNRestTestCase {
+
+
+    private Map<String, Float[]> getL2TestData() {
+        Map<String, Float[]> data = new HashMap<>();
+        data.put("1", new Float[]{6.0f, 6.0f});
```
vamshin, line 77: How about we have a document without the vector and then assert that particular doc comes at the end of the result? Same for both ```CosineTestData``` as well.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
VijayanB: I added so that docExists is synchronized. Since it is just one instruction, i think i can remove this.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
VijayanB: I added thread safe to prevent any race condition based on feedback, since docExists setting is happening in one place and reading is happening on another place. If there is not threat with race condition then definitely we can remove this. At the end, it comes to do we need to make this instance thread safe or not? What do you think?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
vamshin: Yes please. I have not seen other places doing this. We can remove this.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
vamshin: Good point. Do you have pointers to any other base classes extending ``` ScriptDocValues<>``` doing synchronization? My understanding is each search request should have its own instance and documents are iterated sequentially so synchronization should not be required. If you see any base class doing this we can definitely consider.

VijayanB: Removed synchronization.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorScriptDocValues.java
VijayanB: Removed

src/test/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/PainlessScriptScoringIT.java
VijayanB: Added test data.

jmazanec15 (approved):
LGTM!

vamshin (approved):
LGTM! Thanks for adding painless support. This will help users to have more customization on the scoring.

Issue #283, opened 2020-12-18
Add nmslib's bit_hamming spaces into plugin


luyuncheng:
As I see #264  that add Hamming distance in custom scoring it is a great functionality. i see there is bit_hamming space [space_bit_hamming](https://github.com/nmslib/nmslib/blob/master/similarity_search/include/factory/space/space_bit_hamming.h) in nmslib. i think may be we could add this into plugin.

i refer to the code [space_bit_hamming](https://github.com/nmslib/nmslib/blob/master/similarity_search/include/factory/space/space_bit_hamming.h) and [space_bit_hamming_test](https://github.com/nmslib/nmslib/blob/master/similarity_search/test/test_space_serial.cc#L231), may be we could add "**SpaceBitVector**" into plugin and support bit_hamming space which is no optimized index.

i also refer to PR: #161 which add no optimized index for "negdotprod", i see the nmslib's python_binding code [python_binding_nmslib](https://github.com/nmslib/nmslib/blob/master/python_bindings/nmslib.cc#L115), may be we could add a "save_data" into plugin and can store index and dataset for "no optimized index". 

so i submit a PR for this.

vamshin:
Hi @luyuncheng,

We have few concerns with the non optimized index for Hamming distance in nmslib. In Elasticsearch we would store serialized graph per segment which means one additional file per ```knn_vector``` field . For non optimized indices like Hamming we will end up having 2 files per segment, one to store the graph and one for the data(elements in the graph). So for large data set,  it is possible to end up with large number of segments which could potentially exhaust file descriptors and run into issues of no available file descriptors. The Pr you mentioned #161 is put into hold for the very same reason. We worked with nmslib team to make optimized index for negative dot product to have one file per segment. We will have a new PR which would enable negative dot product with optimized index.

There are couple of suggestions
1. Enable optimized index support for Hamming in nmslib and then incorporated the changes in k-NN plugin
2. Make use of custom scoring feature for Hamming. 

How about you start with the 2nd approach and let us know if you see any performance concerns with custom scoring for Hamming?  

 We could then take a call about having optimized/non optimized Hamming index? Till then we would like to keep your PR(https://github.com/opendistro-for-elasticsearch/k-NN/pull/284) for hamming support on hold.

PR #161, opened 2020-07-07
Support dot product similarity measure and change scoring function (Backward Compatible)

chenqi0805:
*Issue #, if available:*
#114 #171 

*Description of changes:*
This PR supports the new space type `negdotprod`. The main issue in supporting the new space type is that nmslib only supports l2 and cosinesimil for optimized index: https://github.com/nmslib/nmslib/blob/32e6e69a574b678da0d37bece8dbe6b1b250b660/python_bindings/README.md#saving-indexes-and-data

while for all other space types Object vector data has to be saved and loaded explicitly from `.dat` file. This PR deals with those new `.dat` files. Changes include

1. JNI code
2. Write footer for .dat file
3. Compound file format
4. Change scoring function into 1-sigmoid(raw KNN similarity score) in KNNWeight to cover the whole real axis.

Note:

This PR should be reviewed after #160 since it is based off the nmslib latest version.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.

PR #284, opened 2020-12-18
[On Hold] Add bit hamming

luyuncheng:
*Issue #283,

*Description of changes:*

As I see #264  that add Hamming distance in custom scoring it is a great functionality. i see there is bit_hamming space [space_bit_hamming](https://github.com/nmslib/nmslib/blob/master/similarity_search/include/factory/space/space_bit_hamming.h) in nmslib. i think we can add this into plugin.

i refer to the code [space_bit_hamming](https://github.com/nmslib/nmslib/blob/master/similarity_search/include/factory/space/space_bit_hamming.h) and [space_bit_hamming_test](https://github.com/nmslib/nmslib/blob/master/similarity_search/test/test_space_serial.cc#L231), may be we could add "**SpaceBitVector**" into plugin and support bit_hamming space which is no optimized index.

i also refer to PR: #161 which add no optimized index for "negdotprod", i see the nmslib's python_binding code [python_binding_nmslib](https://github.com/nmslib/nmslib/blob/master/python_bindings/nmslib.cc#L115), may be we could add a "save_data" into plugin and can store index and dataset for "no optimized index". 

so, i submit this PR.

### DIFF
--- a/jni/include/com_amazon_opendistroforelasticsearch_knn_index_v206_KNNIndex.h
+++ b/jni/include/com_amazon_opendistroforelasticsearch_knn_index_v206_KNNIndex.h
@@ -10,10 +10,26 @@ extern "C" {
 /*
  * Class:     com_amazon_opendistroforelasticsearch_knn_index_v206_KNNIndex
  * Method:    saveIndex
- * Signature: ([I[[FLjava/lang/String;[Ljava/lang/String;Ljava/lang/String;)V
+ * Signature: ([I[[FLjava/lang/String;[Ljava/lang/String;Ljava/lang/String;Z)V
  */
 JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v206_KNNIndex_saveIndex
-  (JNIEnv *, jclass, jintArray, jobjectArray, jstring, jobjectArray, jstring);
+  (JNIEnv *, jclass, jintArray, jobjectArray, jstring, jobjectArray, jstring, jboolean);
+
+/*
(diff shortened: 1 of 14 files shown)


jmazanec15:
Hi @luyuncheng,

Thanks for the PR! I will take a look some time early next week. My concern about the "non-optimized" indices is that we will end up storing two different files for the HNSW structure in the Lucene index. I need to think about the implications of doing this some more.

luyuncheng:
> Hi @luyuncheng,
> 
> Thanks for the PR! I will take a look some time early next week. My concern about the "non-optimized" indices is that we will end up storing two different files for the HNSW structure in the Lucene index. I need to think about the implications of doing this some more.

#161 in this PR, it Write footer for .dat file. I reuse this as "non-optimized" data storage.

As i checked the nmslib's code, When we add "non-optimized" data storage, K-NN plugin can support all nmslib's Space Types

jmazanec15:
@luyuncheng  Apologies, have not gotten a chance to look at this yet. Will review in January.

luyuncheng:
Previously, this PR using string to parse the bit vector, this may occupy too much memory!
As i checked the code: [space_bit_hamming.h](https://github.com/nmslib/nmslib/blob/master/similarity_search/include/space/space_bit_vector.h) in nmslib, i think use a int32 array to represent a bit vector is better. so, add modified some code to use vector<int32> as bit vector

vamshin:
This Pr is put on hold. Please refer https://github.com/opendistro-for-elasticsearch/k-NN/issues/283#issuecomment-755598088

Issue #225, opened 2020-09-15
Adding support for FAISS


raman-r-4978:
Do you guys have any plans to support faiss other than nmslib in future?

Few issues I have encountered while using nmslib is,
- It doesn't support mmap, leads to massive memory consumptions while reading index graphs.
- Nowadays, the most common vector size for NLP is 768 or 1024, with that adding ~1M vectors to the nmslib index takes a very very long time to build graphs compared to the faiss IVFFlat type of indexes
- Speaking of time, considering graph merges adds a whole additional complications to this.

Since faiss has a significant solutions to handle these issues, I would be happy to have both of them integrated into this plugin.

Attaching an [ES issue thread](https://github.com/elastic/elasticsearch/issues/42326) that you might be interested in.

I have also created Java bindings for faiss which can be found [here](https://github.com/RamanRajarathinam/JFaiss-CPU).

Hope it helps

vamshin:
Hi @RamanRajarathinam,

We are in plans to support FAISS. Nothing concrete yet. Will keep this thread open as a Feature request. Based on the community feedback we could prioritize the possibility of this feature. Those who end up on this thread looking for FAISS support please +1 this thread.

Kavan72:
+1

YashalKanungo:
+1

hiro-v:
+1

walker313504:
+1

greav:
+1

jmazanec15:
As an update, we are working to add faiss support to the plugin. We recently received a contribution to add the library and its HNSW implementation. Because we do not see improvement with faiss's HNSW versus nmslib's, we have decided to incorporate other faiss methods before releasing. We will build off of that contribution in faiss-support branch. We are looking into adding functionality for inverted file systems, product quantization, as well as composite indices. Because these methods require training, the implementation is a little more complex. In the coming weeks, we will publish an RFC. In the meantime, please feel free to "+1" or mention a specific feature from faiss you would like to have supported.

alwc:
+1

luyuncheng:
+1, As  [ml-supervised-workflow](https://www.elastic.co/guide/en/machine-learning/7.11/ml-supervised-workflow.html) shows. may be we can use some workflow in faiss training

jmazanec15:
@luyuncheng That is a Elastic commercial feature, so we cannot use that.

[...]

PR #285, opened 2020-12-25
Adding support for FAISS JNI

luyuncheng:
*As Issue #225 mentioned. It is possible to support faiss as future.

But i do not implement the codec services, because i want to ask for your opinions about this part.

I prefer to use the same interface(**`saveIndex`, `init`, `queryIndex`**) as same as nmslib's code with the same input parameter, this would i have more compatibility for knn plugin. so i PR this using the same JNI interface.
i think with the same interface we can support nmslib and faiss in knn plugin at the same time.

*Description of changes:*
1. This PR is support `Faiss` index and Query with the same interface in jni code.
2. Add `index.knn.knnEngine` to distinguish faiss and nmslib.
3. When Write an Index in DocValuesConsumer, using `index.knn.knnEngine` to load the explicit Library.
4. When Query an Index in KNNWeigh, using `index.knn.knnEngine` to load the explicit Library.
5. In `KNNJNIFaissTests` tests, it can do the same logic as `KNNIndex` and `KNNJNITests` 
6. It almost PASS all `integTest` when using Faiss as Library
7. We can modified interface code as less as mush.
8. To build this, we need to [`build faiss`](https://github.com/facebookresearch/faiss/blob/master/INSTALL.md) firstly which depends on some external Libs like OpenMP,BLAS etc. 

I think this PR, there is a lot of details need to be discussion. 
PS. Is there any possible to open a faiss branch?


jmazanec15:
Hi @luyuncheng , thanks for making this PR! This is a very big change, so it will take some time to review. I will start looking at it in January.

jniFaiss/CMakeLists.txt
```diff
@@ -0,0 +1,144 @@
+cmake_minimum_required(VERSION 2.8)
```
jmazanec15, line 1: Design question: Should we have separate JNI libraries for FAISS and nmslib, or should they be contained in one?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -151,12 +153,22 @@ private void onRemoval(RemovalNotification<String, KNNIndexCacheEntry> removalNo
      */
     public KNNIndex getIndex(String key, final String indexName) {
         try {
+            //TODO if Type Not consistent
```
jmazanec15, line 156: What do you mean by this?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
```diff
@@ -83,6 +85,7 @@
     /**
      * Default setting values
      */
+    public static final String INDEX_KNN_DEFAULT_ENGINE = "Faiss"; // nmslib, faiss
```
jmazanec15: I think we should keep nmslib as the default engine.

jmazanec15 (commented):
I think we will review the PR first and then check into a develop branch for further testing.

Maintaining the same interface makes sense.

I added a preliminary review. I will continue to review as I think more about it.

A couple comments:
1. For FAISS submodule, we should align it with a release. It appears to be aligned to an arbitrary commit
2. In general, I do not think it is a good idea to check a binary in (libfaiss.a). What is the reason for doing this?

I need to think more on best way to support 2 libraries from the user perspective. I will get back on this.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
luyuncheng: when cached an Index, should we verify which engine for this index, and then load the exactly library?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
luyuncheng: Yes, you are right. this is just for integration test can cover the `faiss` engine code.

jniFaiss/CMakeLists.txt
luyuncheng: i am not sure which is better.
At this, i separate libraries just to elaborate this `faiss engine` can work with knn-plugin.
may be one jni interface can make jni code more simple to maintain

build.gradle
```diff
@@ -120,21 +120,37 @@ loggerUsageCheck.enabled = false
...
+task buildJniFaissLib(type:Exec) {
+    dependsOn cmakeJniFaissLib
+    workingDir 'jniFaiss'
+    commandLine 'make'
+}
+task buildJniLib() {
+    dependsOn cmakeJniLib
+    dependsOn buildJniNmsLib
+    dependsOn buildJniFaissLib
+}
```
vamshin: You could remove  ```dependsOn cmakeJniLib``` and just rely on ```dependsOn buildJniNmsLib``` ```dependsOn buildJniFaissLib``` as they internally depends on ```cmakeJniNmsLib``` and ```cmakeJniFaissLib```

jniFaiss/include/com_amazon_opendistroforelasticsearch_knn_index_faiss_KNNFIndex.h
```diff
@@ -0,0 +1,53 @@
...
+#define _Included_com_amazon_opendistroforelasticsearch_knn_index_faiss_KNNFIndex
+#ifdef __cplusplus
+extern "C" {
+#endif
+/*
+ * Class:     com_amazon_opendistroforelasticsearch_knn_index_faiss_KNNFIndex
+ * Method:    saveIndex
+ * Signature: ([I[[FLjava/lang/String;[Ljava/lang/String;Ljava/lang/String;)V
+ */
+JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_faiss_KNNFIndex_saveIndex
```
vamshin: Is faiss library backward compatible? 
We should have faiss library version in the function names to support backward compatibility issues if any arises in the future? You could refer nmslib apis.

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_KNNFIndex.cpp
```diff
@@ -0,0 +1,272 @@
...
+        catch_cpp_exception_and_throw_java(env);	
+	}
+	return NULL;
+}
+
+/**
+ * When autoclose class do close, then delete the pointer
+ * Method GC pointer
+ */
+JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_faiss_KNNFIndex_gc
```
vamshin: We should run memory leak tests for faiss jni code. @jmazanec15 ran some memory leak tests for nmslib, we could probably ran it on the faiss branch once we create. Lets see if we can test it as part of this PR or atleast we can create an issue to track all TODOs later.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/util/NmsLibVersion.java
```diff
@@ -25,9 +28,17 @@
         public String indexLibraryVersion() {
             return "KNNIndexV2_0_6";
         }
+    },
+    VFaiss("Faiss") {
```
vamshin: Having Faiss part of ```NmsLibVersion``` class would add confusion. We should probably have dedicated enum class for Faiss versions? Intension of this class is to hold different versions of same library and refer them from this class incase we happen to maintain more than one version of the library because of compatability issues.

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNJNIFaissTests.java
```diff
@@ -0,0 +1,177 @@
...
+import org.apache.lucene.store.FilterDirectory;
+
+import java.nio.file.Paths;
+import java.security.AccessController;
+import java.security.PrivilegedAction;
+import java.util.Arrays;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class KNNJNIFaissTests extends KNNTestCase {
```
vamshin, line 19: We should also have some tests that create both nmslib and faiss indices and perform search in the same test and assert right values are retrieved. This is to verify we are not mixing up between library calls.

src/main/plugin-metadata/plugin-security.policy
```diff
@@ -1,3 +1,4 @@
 grant {
     permission java.lang.RuntimePermission "loadLibrary.KNNIndexV2_0_6";
+    permission java.lang.RuntimePermission "loadLibrary.KNNIndexVFaiss";
```
vamshin: We should add version information similar to NMSlib to keep it consistent and provides ability to load certain version of the library.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/faiss/KNNFIndex.java
```diff
@@ -0,0 +1,138 @@
+package com.amazon.opendistroforelasticsearch.knn.index.faiss;
```
vamshin: package name should have version information similar to nmslib.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -98,16 +99,29 @@ public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer)
...
+                        new PrivilegedAction<Void>() {
+                            public Void run() {
+                                KNNIndex.saveIndex(pair.docs, pair.vectors, tempIndexPath, algoParams, spaceType);
+                                return null;
+                            }
                         }
-                    }
-            );
+                );
+            }
```
vamshin: How about we move if else check inside the ```AccessController.doPrivileged```. Something like below example

```
                AccessController.doPrivileged(
                        new PrivilegedAction<Void>() {
                            public Void run() {
                                if(knnindex)
                                      KNNIndex.saveIndex(pair.docs, pair.vectors, tempIndexPath, algoParams, spaceType);
                                if(faissindex)
                                     KNNFIndex.saveIndex(pair.docs, pair.vectors, tempIndexPath, algoParams, spaceType);      
                               return null;
                            }
      ```

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java [hunk 12]
```diff
@@ -97,11 +100,21 @@ public Scorer scorer(LeafReaderContext context) throws IOException {
...
 
             Path indexPath = PathUtils.get(directory, hnswFiles.get(0));
-            final KNNIndex index = knnIndexCache.getIndex(indexPath.toString(), knnQuery.getIndexName());
-            final KNNQueryResult[] results = index.queryIndex(
-                    knnQuery.getQueryVector(),
-                    knnQuery.getK()
-            );
+            final KNNQueryResult[] results;
+
+            if (fieldAttributes.containsValue(NmsLibVersion.V206.getBuildVersion())) {
```
vamshin: How about we pull out the methods in KNNFIndex and KNNIndex to interface so that we could refer with one object to keep it more cleaner? Something like

```
KNNIndex index;
if (fieldAttributes.containsValue(NmsLibVersion.V206.getBuildVersion())) {
       index = knnIndexCache.getIndex(indexPath.toString(), knnQuery.getIndexName());
} else {
       index = knnIndexCache.getFIndex(indexPath.toString(), knnQuery.getIndexName());
}
results = index.queryIndex(knnQuery.getQueryVector(), knnQuery.getK());



```

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_KNNFIndex.cpp
```diff
@@ -0,0 +1,272 @@
+#include "com_amazon_opendistroforelasticsearch_knn_index_faiss_KNNFIndex.h"
```
vamshin: This class has core logic and would need some insights into way faiss library operates to review the code. I plan to visit this code once I have understanding of faiss interface.

jniFaiss/CMakeLists.txt
vamshin: I feel having seperate JNI would be more cleaner and easy to abstract out the underlying business logic to dedicated files.  I like the current approach.

vamshin (commented):
Few questions

1. Does faiss support all the library functions nmslib supports. We are primarily focusing on
L2, cosine, L1, negative dot product. Accordingly we may need to have relevant validations and throw the exceptions with proper message.

2. Can we confirm From Elasticsearch point of view following work? Manual test is fine for now and ideally we would prefer to have integration tests. 
    1. A nmslib index should be possible to reindex into faiss index and vice versa
    2. Forcemerge calls work on faiss index (recall should be about same)
    3. snapshot/restore work

build.gradle
luyuncheng: Thanks, I fixed it

jniFaiss/include/com_amazon_opendistroforelasticsearch_knn_index_faiss_KNNFIndex.h
luyuncheng: Yes, your are right. i will add the version number into it.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/util/NmsLibVersion.java
luyuncheng: Nice idea! i'll do it

luyuncheng:
> 2\. Can we confirm From Elasticsearch point of view following work? Manual test is fine for now and ideally we would prefer to have integration tests. @vamshin 

I added some code for the integration tests in CI. and it pass all the tests. 
the default faiss library build with CPU part only. which reference from:[ general-compilation-instructions ](https://github.com/facebookresearch/faiss/blob/master/INSTALL.md#general-compilation-instructions)

Next, i'll do following things:
- [x] add faiss library version in the function names
- [x] add some tests for both faiss and nmslib library

add faiss library version in the function names with following versions:

```
public enum NmsLibVersion {
    VNMSLIB_208("NMSLIB_208"){
        @Override
        public String indexLibraryVersion() {
            return "KNNIndex_NMSLIB_V2_0_8";
        }
    },
    VFAISS_164("FAISS_164") {
        @Override
        public String indexLibraryVersion() {
            return "KNNIndex_FAISS_V1_6_4";
        }
    };
}
```

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java
luyuncheng: this is may make the log of changes in KNNIndex. In KNNIndex class, it would load the specific library, if we want to pull out the methods `getIndex`, we need an abstraction of KNNIndex and verify the engine we are using.
Let me think about it how to do it.

src/main/plugin-metadata/plugin-security.policy
luyuncheng: How about make 
nmslib: KNNIndex_NMSLIB_V2_0_6, 
faiss: KNNIndex_FAISS_V1_6_4 
like this:
```
public enum NmsLibVersion {
    VNMSLIB_208("NMSLIB_208"){
        @Override
        public String indexLibraryVersion() {
            return "KNNIndex_NMSLIB_V2_0_8";
        }
    },
    VFAISS_164("FAISS_164") {
        @Override
        public String indexLibraryVersion() {
            return "KNNIndex_FAISS_V1_6_4";
        }
    };
}
```
or do you have any idea to distinguish the name between nmslib and faiss library.

With this, we may need to change the package 
`com.amazon.opendistroforelasticsearch.knn.index.v208` to `com.amazon.opendistroforelasticsearch.knn.index.nmslib.v208` and `com.amazon.opendistroforelasticsearch.knn.index.faiss` to `com.amazon.opendistroforelasticsearch.knn.index.faiss.v164`

Load Library:
```
    permission java.lang.RuntimePermission "loadLibrary.KNNIndex_NMSLIB_V2_0_8";
    permission java.lang.RuntimePermission "loadLibrary.KNNIndex_FAISS_V1_6_4";
```

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/util/NmsLibVersion.java
luyuncheng: How about make 
nmslib: KNNIndex_NMSLIB_V2_0_6, 
faiss: KNNIndex_FAISS_V1_6_4 
like this:
```
public enum NmsLibVersion {
    VNMSLIB_208("NMSLIB_208"){
        @Override
        public String indexLibraryVersion() {
            return "KNNIndex_NMSLIB_V2_0_8";
        }
    },
    VFAISS_164("FAISS_164") {
        @Override
        public String indexLibraryVersion() {
            return "KNNIndex_FAISS_V1_6_4";
        }
    };
}
```
or do you have any idea to distinguish the name between nmslib and faiss library.

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_KNNFIndex.cpp
luyuncheng: is there any tools to detect memory leak in JNI code? 
when i submit the faiss jni code, i have been writed a demo with `valgrind` to detect memory leak. but i do not know how to detect it from java test code.

vamshin: Good point. We used ```valgrind``` to detect memory leaks. Last time I remember we build C++ class to call the functions directly and then build the artifact for running ```valgrind```. We could probably help here.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/util/NmsLibVersion.java
vamshin: We could have new enum class for FAISS that just holds FAISS versions.  Example

```
public enum FAISSLibVersion {

    /**
     * Latest available faiss version
     */
    V_164("164"){
        @Override
        public String indexLibraryVersion() {
            return "KNNIndexV2_0_11";
        }
    };

    public static final FAISSLibVersion LATEST = V164;

    public String buildVersion;

    FAISSLibVersion(String buildVersion) {
        this.buildVersion = buildVersion;
    }

    /**
     * FAISS library version used by the KNN codec
     * @return name
     */
    public abstract String indexLibraryVersion();
}
```

.github/workflows/CI.yml
```diff
@@ -4,6 +4,7 @@ on:
     branches:
       - master
       - opendistro-*
+      - faiss*
```
jmazanec15: I dont think we need to add this branch to CI action, given that it is a development branch. Please remove. Once review finishes, we will check into faiss branch and develop/test from there. Then, once we are ready, we can create another PR to master.

.github/workflows/test-workflow.yml
```diff
@@ -5,10 +5,12 @@ on:
     branches:
       - master
       - opendistro-*
+      - faiss*
   push:
     branches:
       - master
       - opendistro-*
+      - faiss*
```
jmazanec15, line 12: Same as above. Please remove.

jniFaiss/CMakeLists.txt
jmazanec15: I think having 2 separate libraries is okay.

jniFaiss/CMakeLists.txt
```diff
@@ -0,0 +1,108 @@
...
+
+project(KNNIndex_FAISS)
```
jmazanec15, line 3: I think this might be better for CMakeLists.txt: https://gist.github.com/jmazanec15/da1e68aed9073c7d4706597ba8ef2087.

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
```diff
@@ -0,0 +1,272 @@
...
+#include "faiss/index_io.h"
+#include "faiss/IndexHNSW.h"
+
+
+using std::string;
+using std::vector;
+
+std::unordered_map<string, faiss::MetricType> mapMetric = {
+        {"l2", faiss::METRIC_L2},
+        {"innerproduct", faiss::METRIC_INNER_PRODUCT}
```
jmazanec15: Doesn't FAISS also have cosine space?

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
```diff
@@ -0,0 +1,272 @@
...
+};
+
+extern "C"
+
+struct JavaException {
```
jmazanec15: Can you add a TODO for future to abstract out common logic from JNI levels? This does not have to be completed in this PR, but we should do it in the future.

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
```diff
@@ -0,0 +1,272 @@
...
+ *
+ */
+JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex_saveIndex
+(JNIEnv* env, jclass cls, jintArray ids, jobjectArray vectors, jstring indexPath, jobjectArray algoParams, jstring spaceType)
+{
+	vector<int64_t> idVector;
+	vector<float>   dataset;
+	vector<string> paramsList;
+	string indexDescription = "HNSW32";
+	faiss::MetricType metric = faiss::METRIC_L2;
```
jmazanec15: I think instead of setting L2 as default, we should just make sure that spaceType is passed in and is valid. Setting L2 as default here could lead to silent failures.

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
```diff
@@ -0,0 +1,272 @@
...
+			for(int j = 0; j < dim; ++j) {
+				dataset.push_back(vector[j]);
+			}
+			env->ReleaseFloatArrayElements(vectorArray, vector, 0);
+		}
+		has_exception_in_stack(env);
+
+		//---- indexPath
+		const char *indexString = env->GetStringUTFChars(indexPath, 0);
+        string indexPathString(indexString);
```
jmazanec15: Indentation in this file looks off. Could you please fix?

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
```diff
@@ -0,0 +1,272 @@
...
+ * Method: saveIndex
+ *
+ */
+JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex_saveIndex
+(JNIEnv* env, jclass cls, jintArray ids, jobjectArray vectors, jstring indexPath, jobjectArray algoParams, jstring spaceType)
+{
+	vector<int64_t> idVector;
+	vector<float>   dataset;
+	vector<string> paramsList;
+	string indexDescription = "HNSW32";
```
jmazanec15: Could you add a TODO here to eventually parametrize this? I think we should be able to support any Faiss index in the future.

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+		//---- algoParams
+		int paramsCount = env->GetArrayLength(algoParams);
+        for (int i=0; i<paramsCount; i++) {
+            jstring param = (jstring) (env->GetObjectArrayElement(algoParams, i));
+            const char *rawString = env->GetStringUTFChars(param, 0);
+            paramsList.push_back(rawString);
+
+            int M = 32;
+            if (sscanf(rawString, "M=%d", &M) == 1) {
```
jmazanec15: Why do we need to sscanf here?

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
```diff
@@ -0,0 +1,272 @@
...
+                        faiss::IndexHNSW* ihp = reinterpret_cast<faiss::IndexHNSW*>(indexWriter.get());
+                        ihp->hnsw.efSearch = efSearch;
+		        }
+		    }
+		}
+
+		//---- Do Index
+		//----- 1. Train
+        if(!indexWriter->is_trained) {
+			//TODO if we use like PQ, we have to train dataset
```
jmazanec15: Yeah I was thinking about this a little bit. It seems like it might make sense to fall back to a flat index if we do not have enough points to train. But we can think about this later.

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
```diff
@@ -0,0 +1,272 @@
...
+
+/**
+ * Method: Global Init
+ *
+ */
+JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex_initLibrary(JNIEnv *, jclass)
+{
+	//set thread 1 cause ES has Search thread
+	//TODO make it different at search and write
+//	omp_set_num_threads(1);
```
jmazanec15: Why is this commented out?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
jmazanec15: I see. I think we should add a different file extension for faiss graphs as opposed to nmslib graphs. Maybe <filename>.faiss_hnsw.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -320,22 +339,52 @@ public KNNIndexCacheEntry loadIndex(String indexPathUrl, String indexName) throw
...
         private final KNNIndex knnIndex;
+        private final KNNFIndex knnFindex;
         private final String indexPathUrl;
         private final String esIndexName;
         private final WatcherHandle<FileWatcher> fileWatcherHandle;
 
         private KNNIndexCacheEntry(final KNNIndex knnIndex, final String indexPathUrl, final String esIndexName,
                                    final WatcherHandle<FileWatcher> fileWatcherHandle) {
             this.knnIndex = knnIndex;
+            this.knnFindex = null;
```
jmazanec15: Seems to me that knnIndex and knnFindex should inherit from some kind of abstract class.

jmazanec15 (commented):
Hi @luyuncheng, could you remove the jniFaiss/external/libfaiss.a? We don't need to ship binaries.

Also, can you update FAISS library commit to their latest release? v1.6.5. Right now, it points to a more recent commit.

.github/workflows/CI.yml
luyuncheng: ok，i'll fix it

jniFaiss/CMakeLists.txt
luyuncheng: nice code! i'll use this one

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
luyuncheng: In Doc [Faiss-indexes](https://github.com/facebookresearch/faiss/wiki/Faiss-indexes) , i added inner product, which means If we have normalized vectors, the inner product becomes cosine similarity. 
And there is some issues talk about this: https://github.com/facebookresearch/faiss/issues/593

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
luyuncheng: i am no sure is it appropriate to set threads as 1 or  reuse the settings:
`public static final String KNN_ALGO_PARAM_INDEX_THREAD_QTY = "knn.algo_param.index_thread_qty";`

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
luyuncheng: > I see. I think we should add a different file extension for faiss graphs as opposed to nmslib graphs. Maybe .faiss_hnsw. 

in the latest commit, i use `knnEngine` name in the index file names like:
```
String hnswFileName = String.format("%s_%s_%s%s", 
    state.segmentInfo.name, 
    knnEngine.getLatestBuildVersion(),
    field.name, 
    KNNCodecUtil.HNSW_EXTENSION);
```

> Seems to me that knnIndex and knnFindex should inherit from some kind of abstract class.

yes, you are right. let me think about how to abstract this.

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
luyuncheng: > Why do we need to sscanf here?

In the KNN plugin we would make algoParam like: `M=32, efSearch=20`, As [faiss index factory docs](https://github.com/facebookresearch/faiss/wiki/The-index-factory#non-exhaustive-search-components) shows, if `M=32` we need use string: `HNSW32` for index Description. so i use sscanf to get M.

i think we need an adapter to match different knnEngine algorithms params. if you have good idea, pls let me know.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/faiss/KNNFIndex.java
luyuncheng: in the latest commits, i add `faiss` version into package and function names.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/util/NmsLibVersion.java
luyuncheng: > We could have new enum class for FAISS that just holds FAISS versions. Example

yes, it is great. i added FAISSLibVersion in the latest commits

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
luyuncheng: L72 just init a default value. at L121 i verify the Metric Space:
```
if(mapMetric.find(spaceTypeString) != mapMetric.end()) {
    metric = mapMetric[spaceTypeString];
}
```

I have a question: if we verify the `metricSpace` in the jni code. this may cause ES shard failed assert when an invalid space param passed in jni code.  should i throw an exception when mapMetric can not find spaceTypeString?

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/KNNJNIFaissTests.java
luyuncheng: yes, i would add tests for mixed engine and test with exceptions case

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
luyuncheng: > Seems to me that knnIndex and knnFindex should inherit from some kind of abstract class.

I used KNNIndex as abstract class. `KNNFaissIndex` and `KNNNmsLibIndex` extends the KNNINdex. 
`KNNIndexCacheEntry` only save the KNNIndex.

if you have any other idea, pls let me know

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java
luyuncheng: > How about we pull out the methods in KNNFIndex and KNNIndex to interface so that we could refer with one object to keep it more cleaner? Something like

in the latest commits, I used KNNIndex as abstract class. `KNNFaissIndex` and `KNNNmsLibIndex` extends the KNNINdex. 
`KNNIndexCacheEntry` only save the KNNIndex.

if you have any other idea, pls let me know

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
jmazanec15: I think that makes sense. Thanks for making the change!

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
jmazanec15: I think it probably makes sense to reuse KNN_ALGO_PARAM_INDEX_THREAD_QTY.

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
jmazanec15: Ah got it. I think makes sense.

jniFaiss/src/com_amazon_opendistroforelasticsearch_knn_index_faiss_v164_KNNFIndex.cpp
jmazanec15: Oh interesting. That makes sense.

jmazanec15 (commented):
Thanks for making this change @luyuncheng!

I think at this point we can check it into a development branch "faiss-support" and I will start working on it/testing it as well.

I created "faiss-support". Could you merge in all the changes to that branch so that we can merge this change in?

luyuncheng:
> Thanks for making this change @luyuncheng!
> 
> I think at this point we can check it into a development branch "faiss-support" and I will start working on it/testing it as well.
> 
> I created "faiss-support". Could you merge in all the changes to that branch so that we can merge this change in?

i have changed the base branch from `master` to `faiss-support`.  
Thanks for reviewing this PR ! @jmazanec15 @vamshin 
Let me know if you need anything else

jmazanec15 (approved):

vamshin (approved):
LGTM! to push to dedicate branch. Once again thanks for the great work @luyuncheng.

We can continue development/testing on top of this code in dedicated branch.

Issue #288, opened 2020-12-29
KNNVectorFieldMapper unable to get algorithm params from settings when built from merge


jmazanec15:
When the KNNVectorFieldMapper is built from a mapper merge, the mapper is unable to read the algorithm parameters that are stored as index settings. Because of this, it falls back to default settings. Because spaceType is one of the parameters, this can lead to incorrect graphs being built for an index. For example, a user may create a `cosinesimil` index and the graphs will be built using the `l2` case.

Relevant lines of code:
1 https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java#L124-L163
2. https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java#L349
3. https://github.com/elastic/elasticsearch/blob/7.10/server/src/main/java/org/elasticsearch/index/mapper/ParametrizedFieldMapper.java#L104

For basic use cases where the mapping does not change after index creation, this bug will not make any impact. However, this bug may impact indices with dynamic mappings.

Reproduction steps:
```
curl -X PUT "localhost:9200/myindex" -H 'Content-Type: application/json' -d'
{
  "settings" : {
      "index": {
        "knn": "true",
        "knn.space_type": "cosinesimil"
      }
  },
  "mappings": {
    "properties": {
      "embedding": {
        "type": "knn_vector",
        "dimension": 2
      },
      "metadata" : {
        "type" : "object",
        "dynamic" : true
      }
    }
  }
}
'

curl -X POST "localhost:9200/myindex/_doc" -H 'Content-Type: application/json' -d'
{
  "metadata": {"test": "object"},
  "embedding":  [1, 2]
}
'

curl -X POST "localhost:9200/myindex/_search" -H 'Content-Type: application/json' -d'
{
  "size" : 1,
  "query": {
    "knn": {
      "embedding": {
        "vector": [3, 4],
        "k": 1
      }
    }
  }
}
'
```

[This change](https://github.com/jmazanec15/k-NN/commit/78e09bc631e84980147542c7d9ff673af5c439b5) should resolve the problem. In the future, we should switch these algorithm parameters from index settings to mapping parameters (tracking this in #282).

Impacted ODFE Version:
1. 1.12
2. 1.11
3. 1.10

Related issues:
[...]

PR #289, opened 2020-12-29
Add builder constructor that takes algo params

jmazanec15:
*Issue #, if available:*
#288 

*Description of changes:*
In this PR, I added a constructor for KNNVectorFieldMapper.Builder that takes the algorithm parameters that it typically reads from settings. This constructor is used during merging of mappers. It allows an existing KNNVectorFieldMapper to pass settings to a KNNVectorFieldMapper.Builder for merge.  It plays a similar role to the `ParametrizedFieldMapper.init` method, which copies mapping parameters from an already existing FieldMapper to the FieldMapper.Builder that will be used for merging. I decided to not override this method because it is more focused on mapping parameters, but I am open to feedback on this.

Verified that with this change, I am unable to reproduce the issue following the steps in #288.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
@@ -101,8 +101,19 @@ public class KNNVectorFieldMapper extends ParametrizedFieldMapper {
                 }, m -> toType(m).dimension);
         private final Parameter<Map<String, String>> meta = Parameter.metaParam();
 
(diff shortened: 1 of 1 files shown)


vamshin (approved):
Good catch. LGTM!

VijayanB:
Is it possible to add integration test or unit test for this scenario? May be as another commit since this is critical.

VijayanB (approved):

jmazanec15:
> Is it possible to add integration test or unit test for this scenario? May be as another commit since this is critical.

Yes, I think we need to add more end to end testing in our integration tests. We can take that as another task.

Issue #290, opened 2020-12-29
Upgrade JNI layer to use nmslib-2.0.11


jmazanec15:
Nmslib 2.0.8 contains several enhancements including:
1. Optimized index support for L1, Linf and Negative Dot Product space
2. ARM compatibility

We need to update our JNI library to use this version

jmazanec15:
Due to issues building python bindings, nmslib release was changed from 2.0.8 to 2.0.11. Although no core changes were made during this switch, for the sake of completeness, we should upgrade to 2.0.11.

PR #291, opened 2020-12-30
Upgrade nmslib to v2.0.8 in JNI library

jmazanec15:
*Issue #, if available:*
#290 

*Description of changes:*
This PR upgrades nmslib to v2.0.8 for our JNI library. Placed in WIP state because I would like to test version upgrade and a few other things.


src/test/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNNCodecTestCase.java
```diff
@@ -153,8 +153,8 @@ public void testMultiFieldsKnnIndex(Codec codec) throws Exception {
         IndexSearcher searcher = new IndexSearcher(reader);
         float score = searcher.search(new KNNQuery("test_vector", new float[] {1.0f, 0.0f, 0.0f}, 1, "dummy"), 10).scoreDocs[0].score;
         float score1 = searcher.search(new KNNQuery("my_vector", new float[] {1.0f, 2.0f}, 1, "dummy"), 10).scoreDocs[0].score;
-        assertEquals(score, 0.1667f, 0.01f);
-        assertEquals(score1, 0.0714f, 0.01f);
+        assertEquals(score, 1.0f/(1 + 25), 0.01f);
```
VijayanB: assertEquals(expected, actual, ... )

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/util/NmsLibVersion.java
```diff
@@ -20,14 +20,14 @@
...
+    V208("208"){
         @Override
         public String indexLibraryVersion() {
-            return "KNNIndexV2_0_6";
+            return "KNNIndexV2_0_8";
         }
     };
 
-    public static final NmsLibVersion LATEST = V206;
+    public static final NmsLibVersion LATEST = V208;
```
VijayanB, line 30: why can't we create a "latest" as enum value, so that we don't have to update this line every time. What do you think?

jni/include/com_amazon_opendistroforelasticsearch_knn_index_nmslib_KNNIndex.h
```diff
@@ -1,50 +1,50 @@
...
   (JNIEnv *, jclass, jlong);
 
 /*
- * Class:     com_amazon_opendistroforelasticsearch_knn_index_v206_KNNIndex
+ * Class:     com_amazon_opendistroforelasticsearch_knn_index_nmslib_KNNIndex
  * Method:    initLibrary
  * Signature: ()V
  */
-JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v206_KNNIndex_initLibrary
+JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_nmslib_KNNIndex_initLibrary
```
vamshin: Lets keep the versioning in the interface so that it would be easy to support backward compatibility when we find one.  lets add v208 similar to v206

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/util/NmsLibVersion.java
jmazanec15: I think we could but Im not sure how it would be better. The purpose of using the versioning is to handle potential backwards compatibility issues if we upgrade to an nmslib version that breaks backwards compatibility. We havent had this issue so far, so we just replace the nmslib version with the new one. 

However, if we end up having this issue, it seems to me, LATEST should just be a pointer to an enum value and enum values should be concrete versions.

src/test/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNNCodecTestCase.java
jmazanec15: Good point. Im trying to fix these as I go. Will update.

jni/include/com_amazon_opendistroforelasticsearch_knn_index_nmslib_KNNIndex.h
jmazanec15: Makes sense. Will update.

jmazanec15:
Ran a few tests. Confirmed that upgrading from library built on v206 to v208 works. Specifically, I created several indices for each space with a library using v206 and then searched them with the library using v208 and I was able to get expected results. 

I removed WIP designation as PR is ready for review.

vamshin (approved):
LGTM! Thanks for upgrading the library to enable ARM support

VijayanB (approved):

Issue #303, opened 2021-01-14
Run knn integ tests in security mode


vamshin:
As part of ODFE mandate every plugin is expected to run integration tests with security plugin enabled

PR #304, opened 2021-01-14
Run KNN integ tests with security plugin enabled

vamshin:
*Issue #, if available:*
https://github.com/opendistro-for-elasticsearch/k-NN/issues/303

*Description of changes:*
As part of ODFE mandate every plugin is expected to run integration tests with security plugin enabled

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -26,3 +26,42 @@ jobs:
       - name: Run build
         run: |
           ./gradlew build
+
+
+
+      - name: Pull and Run Docker for security tests
+        run: |
+          plugin=`ls build/distributions/*.zip`
+          version=`echo $plugin|awk -F- '{print $4}'| cut -d. -f 1-3`
+          plugin_version=`echo $plugin|awk -F- '{print $4}'| cut -d. -f 1-4`
+          echo $version
+          cd ..
+          if docker pull opendistroforelasticsearch/opendistroforelasticsearch:$version
+          then
+            echo "FROM opendistroforelasticsearch/opendistroforelasticsearch:$version" >> Dockerfile
+            echo "RUN if [ -d /usr/share/elasticsearch/plugins/opendistro-knn ]; then /usr/share/elasticsearch/bin/elasticsearch-plugin remove opendistro-knn; fi" >> Dockerfile
+            echo "ADD k-NN/build/distributions/opendistro-knn-$plugin_version.zip /tmp/" >> Dockerfile
+            echo "RUN /usr/share/elasticsearch/bin/elasticsearch-plugin install --batch file:/tmp/opendistro-knn-$plugin_version.zip" >> Dockerfile
+            docker build -t odfe-knn:test .
+            echo "imagePresent=true" >> $GITHUB_ENV
+          else
+            echo "imagePresent=false" >> $GITHUB_ENV
+          fi
+      - name: Run Docker Image
+        if: env.imagePresent == 'true'
+        run: |
+          cd ..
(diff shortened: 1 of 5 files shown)


.github/workflows/CI.yml
```diff
@@ -26,3 +26,44 @@ jobs:
...
           ./gradlew build
+
+      - name: Pull and Run Docker for security tests
+        run: |
+          plugin=`ls build/distributions/*.zip`
+          version=`echo $plugin|awk -F- '{print $4}'| cut -d. -f 1-3`
+          plugin_version=`echo $plugin|awk -F- '{print $4}'| cut -d. -f 1-4`
+          echo $version
+          cd ..
+          if docker pull opendistroforelasticsearch/opendistroforelasticsearch:$version
```
jmazanec15, line 37: What will happen during version upgrade?

build.gradle
```diff
@@ -150,6 +150,10 @@ integTest {
     // allows integration test classes to access test resource from project root path
     systemProperty('project.root', project.rootDir.absolutePath)
 
+    systemProperty "https", System.getProperty("https")
+    systemProperty "user", System.getProperty("user")
```
jmazanec15, line 154: Will these properties change how we run "./gradlew build"?

src/main/plugin-metadata/plugin-security.policy
```diff
@@ -1,3 +1,4 @@
 grant {
     permission java.lang.RuntimePermission "loadLibrary.KNNIndexV2_0_11";
+    permission java.net.SocketPermission "*", "connect,resolve";
```
jmazanec15, line 3: I see this changing the security policy. How will this impact k-NN during use? Do other plugins do this?

src/test/java/com/amazon/opendistroforelasticsearch/knn/ODFERestTestCase.java
```diff
@@ -0,0 +1,159 @@
...
+import org.elasticsearch.common.xcontent.NamedXContentRegistry;
+import org.elasticsearch.common.xcontent.XContentParser;
+import org.elasticsearch.common.xcontent.XContentType;
+import org.elasticsearch.test.rest.ESRestTestCase;
+import org.junit.After;
+
+/**
+ * ODFE integration test base class to support both security disabled and enabled ODFE cluster.
+ */
+public abstract class ODFERestTestCase extends ESRestTestCase {
```
jmazanec15, line 51: Is there no way to add ODFERestTestCase as an external dependency instead of importing the code?

build.gradle
vamshin: No they do not have any impact.  These will be used only when ```-Dhttps=true``` is passed along with ./gradlew build

src/main/plugin-metadata/plugin-security.policy
vamshin: We needed this to run security tests. Yes other plugins whitelist these.

src/test/java/com/amazon/opendistroforelasticsearch/knn/ODFERestTestCase.java
vamshin: Good point. At this point we do not have external library. For now to unblock I think its ok to have this code.

.github/workflows/CI.yml
vamshin: Good point. This is tricky. For Elasticsearch version upgrade, this would fail as we will not be having the new docker image with Security plugin and Core  Elasticsearch. This is kind of expected and we should ignore the security tests failure during the upgrade and ensure these are rerun once the docker image with upgraded version is released.

jmazanec15 (approved):

VijayanB (approved):

PR #312, opened 2021-01-30
Upgrade odfe version to 1.13.0

VijayanB:
Upgrade to ES 7.10.2


jmazanec15 (approved):
LGTM!

vamshin (commented):
You might want to change here as well https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/gradle.properties#L16.

Even though this field is not used but just to keep it consistent.

VijayanB:
> You might want to change here as well https://github.com/opendistro-for-elasticsearch/k-NN/blob/master/gradle.properties#L16.
> 
> Even though this field is not used but just to keep it consistent.

Ack.

vamshin (approved):
LGTM! Thanks for the changes.

PR #317, opened 2021-02-05
Change mode for jni arrays release to prevent unnecessary copy backs

jmazanec15:
*Issue #, if available:*

*Description of changes:*
Small improvement to JNI layer. For background, once native code no longer needs access to one of the Java components, release has to be called to inform VM that the native code does not need to access it any longer. For primitive arrays, the JNI will either pin the Java array in memory and give the native code a handle to it, or it will create a copy. 

For arrays, the last argument is mode. mode has no impact if the array is not a copy. If it is a copy, mode determines how the release should be handled. We pass "0" which will copy the copy back and then free the buffer. Because we do not change the array, this is unnecessary; instead we can pass JNI_ABORT, which just frees the buffer. 

For more information, please refer to https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#Release_PrimitiveType_ArrayElements_routines.


VijayanB (approved):

vamshin (approved):
LGTM! Thanks for fixing this.

spoonnotfound:
As far as I know, GetPrimitiveArrayCritical method is faster than GetXXXArrayElements. So why not use the primitive methods?

jmazanec15:
Hi @flyingpot , I think that GetPrimitiveArrayCritical may be a little bit more of a risky operation, but worth exploring. I will open an issue. We can discuss there

Hamlin-Li:
Yes, critical function (e.g. GetPrimitiveArrayCritical ) currently is a bit risky for most of GC in hotspot, it will get better for G1 at least after https://bugs.openjdk.java.net/browse/JDK-8276094.

PR #318, opened 2021-02-05
Update minimum score to 0.

VijayanB:
Since scores will be >=0, update min score to Zero from FLOAT.MIN_VALUE
to be consistent with elastic's score representation.


jmazanec15 (approved):
LGTM!

vamshin (approved):
LGTM! Thanks for the fix

PR #319, opened 2021-02-05
Abstract out components and clean up JNI layer to more easily support additional libraries

jmazanec15:
*Issue #, if available:*

*Description of changes:*
This PR does a bit of refactoring of the jni library so that code can be re-used when adding support for other libraries. It takes out common code related to exception handling and JNI translations into a separate jni_util module.

Additionally, it cleans up a few other things such as naming, style, exception checking, clang-tidy warnings. For naming of utility functions, I decided to follow Google's conventions here: https://google.github.io/styleguide/cppguide.html#Function_Names. Overall, this change may slightly improve performance, but not significantly.

### DIFF
--- a/jni/src/com_amazon_opendistroforelasticsearch_knn_index_v2011_KNNIndex.cpp
+++ b/jni/src/com_amazon_opendistroforelasticsearch_knn_index_v2011_KNNIndex.cpp
@@ -71,15 +71,9 @@ JNIEXPORT void JNICALL Java_com_amazon_opendistroforelasticsearch_knn_index_v201
         env->ReleaseIntArrayElements(ids, objectIds, 0);
         index = MethodFactoryRegistry<float>::Instance().CreateMethod(false, "hnsw", spaceTypeCppString, *space, dataset);
 
-        int paramsCount = env->GetArrayLength(algoParams);
-        vector<string> paramsList;
-        string paramString;
-        for (int i=0; i<paramsCount; i++) {
-            paramString = getStringJenv(env, (jstring)(env->GetObjectArrayElement(algoParams, i)));
-            paramsList.push_back(paramString);
-        }
-
+        auto paramsList = getVectorOfStrings(env, algoParams);
         string indexPathCppString = getStringJenv(env, indexPath);
+
         index->CreateIndex(AnyParams(paramsList));
         index->SaveIndex(indexPathCppString);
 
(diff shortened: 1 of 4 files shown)


jni/src/com_amazon_opendistroforelasticsearch_knn_index_v2011_KNNIndex.cpp
```diff
@@ -38,196 +39,135 @@ using similarity::Object;
...
-    try {
-        IndexWrapper *indexWrapper = reinterpret_cast<IndexWrapper*>(indexPointer);
-        has_exception_in_stack(env);
-        delete indexWrapper;
-        has_exception_in_stack(env);
-    }
-    catch (...) {
-        catch_cpp_exception_and_throw_java(env);
-    }
+    auto *indexWrapper = reinterpret_cast<IndexWrapper*>(indexPointer);
```
vamshin, line 166: Minor: Could there be possibility of exception here?

jni/src/com_amazon_opendistroforelasticsearch_knn_index_v2011_KNNIndex.cpp
```diff
@@ -38,196 +39,135 @@ using similarity::Object;
...
-        has_exception_in_stack(env);
 
         return (jlong) indexWrapper;
     }
     // nmslib seems to throw std::runtime_error if the index cannot be read (which
     // is the only known failure mode for init()).
     catch (...) {
-        if (indexWrapper) delete indexWrapper;
-        catch_cpp_exception_and_throw_java(env);
+        delete indexWrapper;
```
vamshin, line 157: why not delete it when its not null like before ```if (indexWrapper) delete indexWrapper;```?
Is it safe to call delete when its null?

Same in other places as well

jmazanec15: Deleting null pointer is no-op: https://en.cppreference.com/w/cpp/language/delete

jni/src/com_amazon_opendistroforelasticsearch_knn_index_v2011_KNNIndex.cpp
jmazanec15: No it will not: https://stackoverflow.com/questions/36916445/can-a-reinterpret-cast-itself-cause-an-exception

jmazanec15:
Closing PR. Will make changes when we merge faiss-support into main.

Issue #171, opened 2020-07-18
Better scoring function in KNN weight


chenqi0805:
Splits from #161 . Our current inverting score function 

https://github.com/opendistro-for-elasticsearch/k-NN/blob/d12cee574428c4b74537916ca3fa5a5cbdcd206c/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java#L113

does not cover raw nmslib knn similarity search score ranging from - maxfloat to + maxfloat. e.g. negative dot product value will not be inverted properly and even introduces singularity at 1. We should replace it with a scoring function that overs the whole real axis and inverts the raw similarity search score accordingly. 

My proposal is to use 1-sigmoid(result.getScore()) = 1 / (1 + exp(result.getScore())). If we want to keep maximum score at 1 for positive raw scores, we can multiply by factor of 2. Comments and suggestions are welcome.

jmazanec15:
Hi @chenqi0805, interesting idea. I have a couple questions about Sigmoid function:

1. What happens when the distance returned by nmslib is high (i.e. 10,000)? Wouldn't this lead to an overflow?

2. What are the performance implications of this approach?

Additionally, another function we could look into for scoring is:
```
x/sqrt(1 + x^2) + 1
```

which would map all distances from 0 to 2.

chenqi0805:
Hi, @jmazanec15 , thx for pointing out the issues.

> 1. What happens when the distance returned by nmslib is high (i.e. 10,000)? Wouldn't this lead to an overflow?

I experimented and found that for x~1000, exp(x) goes to inf and exp(-x) goes to zero. So the asymptotic behavior of sigmoid is not ideal for machine precision.

> 2. What are the performance implications of this approach?
> Additionally, another function we could look into for scoring is:

> x/sqrt(1 + x^2) + 1
> which would map all distances from 0 to 2.

I doubt the time complexity is O(log x) for exponentiation. But I like your proposal except that it should be `-x/sqrt(1 + x^2) + 1` for inverting the score. Here are some benchmarks

```
class Scratch {
  public static double func(double x) {
    return -x/Math.sqrt(1 + x*x) + 1;
  }
    public static void main(String[] args) {
        double x = -100;
        System.out.println(func(x));

        x = -1000;
        System.out.println(func(x));

        x = -10000;
        System.out.println(func(x));

      x = -50000;
      System.out.println(func(x));

      x = -100000;
      System.out.println(func(x));

[...]

Issue #114, opened 2020-05-18
Support for inner product similarity measures (dot product)


vamshin:
https://github.com/nmslib/nmslib/blob/master/manual/spaces.md#inner-product-spaces.

Feel free to +1 if you land up here looking for this feature :)

chenqi0805:
By experimenting with `negdotprod` I find a bug in load index:

```
Check failed: The number of stored elements 3 doesn't match the number of data points ! Did you forget to re-load data?
java.lang.Exception: Check failed: The number of stored elements 3 doesn't match the number of data points ! Did you forget to re-load data?
	at __randomizedtesting.SeedInfo.seed([45BD2309FB22BD07:E1AD9DFC2D566FE4]:0)
	at com.amazon.opendistroforelasticsearch.knn.index.v1736.KNNIndex.init(Native Method)
	at com.amazon.opendistroforelasticsearch.knn.index.v1736.KNNIndex.loadIndex(KNNIndex.java:123)
	at com.amazon.opendistroforelasticsearch.knn.index.KNNJNITests.testAddAndQueryHnswIndexNegDotProd(KNNJNITests.java:192)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:564)
```

This issue is supposed to be fixed in nmslib 1.8 according to https://github.com/nmslib/nmslib/issues/356

I propose an upgrade of the external library before tackling this. Thought? @vamshin @jmazanec15

vamshin:
@chenqi0805  sounds good to me. We need to probably upgrade the library version first. Also do we have data points on the backward compatibility.  Does nmslib 1.8 version can  load/query indices created on nmslib 1.7 version?

chenqi0805:
> Also do we have data points on the backward compatibility. Does nmslib 1.8 version can load/query indices created on nmslib 1.7 version?

@vamshin Need further investigation. Overall upgrade is worth a separate issue.

vamshin:
Sounds good. Created issue https://github.com/opendistro-for-elasticsearch/k-NN/issues/128

[...]

PR #324, opened 2021-02-22
Add support for inner product in ANN, custom scoring and painless

chenqi0805:
*Issue #, if available:*
#114 #171 

*Description of changes:*
This PR
(1) Add negdotprod similarity measure
(2) extend KNN scorer to cover negative similarity measures.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java
@@ -110,7 +110,7 @@ public class KNNWeight extends Weight {
              * neighbors we are inverting the scores.
              */
             Map<Integer, Float> scores = Arrays.stream(results).collect(
-                    Collectors.toMap(result -> result.getId(), result -> score(result.getScore())));
+                    Collectors.toMap(result -> result.getId(), result -> normalizeScore(result.getScore())));
             int maxDoc = Collections.max(scores.keySet()) + 1;
             DocIdSetBuilder docIdSetBuilder = new DocIdSetBuilder(maxDoc);
             DocIdSetBuilder.BulkAdder setAdder = docIdSetBuilder.grow(maxDoc);
@@ -124,12 +124,10 @@ public class KNNWeight extends Weight {
         return true;
     }
 
-    public static float score(float score) {
-        if (score >= 0) {
+    public static float normalizeScore(float score) {
+        if (score >= 0)
             return 1 / (1 + score);
-        } else {
-            return 2 + 1 / (score - 1);
-        }
+        return 2 + 1 / (score - 1);
     }
 }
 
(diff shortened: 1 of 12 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java
```diff
@@ -123,5 +123,13 @@ public Scorer scorer(LeafReaderContext context) throws IOException {
     public boolean isCacheable(LeafReaderContext context) {
         return true;
     }
+
+    public static float scoreFunc(float score) {
```
jmazanec15: nit: Could you just name this score?

jmazanec15:
@chenqi0805 Could you change user exposed names to InnerProduct? I think that would make more sense. For calls to nmslib, we could translate inner_product -> negdotprod. What do you think?

chenqi0805:
@jmazanec15 Sure. ~I will check the places where it is user exposed~. Will open a separate issue to track the translation.

jmazanec15 (approved):
This looks good to me @chenqi0805! Thanks for making this change!

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java
```diff
@@ -123,5 +123,13 @@ public Scorer scorer(LeafReaderContext context) throws IOException {
     public boolean isCacheable(LeafReaderContext context) {
         return true;
     }
+
+    public static float score(float score) {
+        if (score >= 0) {
+            return 1 / (1 + score);
+        } else {
+            return 2 + 1 / (score - 1);
```
VijayanB:
```
private static float normalizeScore(float score) {
    if (score >= 0)
         return 1 / ( 1 + score );
   return 2 + ( 1 / (score - 1) );
   ```

VijayanB: Thanks for making changes. 
1. Do we need this to be public? 
2. Shall we add parenthesis around formula? For better readability, i am thinking we don't want reader to remember the precedence logic to understand the formula. 
What do you think? @chenqi0805

chenqi0805: > 1. Do we need this to be public?

Sorry, I missed the access modifier in your previous comment. The reason I maintained it public is to reuse it in `KNNScoringSpace`. If made private, we will need to duplicate the logic there. Let me know what you think.

> 2. Shall we add parenthesis around formula? For better readability, i am thinking we don't want reader to remember the precedence logic to understand the formula.
What do you think? @chenqi0805

Sure. Makes sense. Will add the parenthesis.

VijayanB (commented):
LGTM! Thanks for making changes!

VijayanB (approved):

Issue #327, opened 2021-02-23
Expose `inner_product` instead of `negdotprod` as space type to users


chenqi0805:
Split from #324 .

chenqi0805:
Closed via #328

PR #328, opened 2021-02-23
Expose inner_product space type, refactoring SpaceTypes.

chenqi0805:
*Issue #, if available:*
#114 #327 

*Description of changes:*
This PR
(1) Refactors SpaceTypes to translate between names in the plugin and param value in nmslib
(1) Exposes inner_product as space_type in ANN, custom scoring
(2) Exposes innerProd in painless.

Note: In painless, I identifies inner_product just as dot product (= - negdotprod) which might be more intuitive for user. Let me know what you think.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java
@@ -26,6 +26,6 @@ public class KNNConstants {
     public static final String LINF = "linf";
     public static final String COSINESIMIL = "cosinesimil";
     public static final String HAMMING_BIT = "hammingbit";
-    public static final String INNER_PROD = "inner_product";
+    public static final String INNER_PROD = "innerproduct";
     public static final String DIMENSION = "dimension";
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java
@@ -26,6 +26,6 @@ public class KNNConstants {
     public static final String LINF = "linf";
     public static final String COSINESIMIL = "cosinesimil";
     public static final String HAMMING_BIT = "hammingbit";
-    public static final String NEGDOTPROD = "negdotprod";
+    public static final String INNER_PROD = "inner_product";
     public static final String DIMENSION = "dimension";
 }
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/SpaceTypes.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/SpaceTypes.java
@@ -15,41 +15,52 @@
 
 package com.amazon.opendistroforelasticsearch.knn.index;
 
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
 /**
  * Enum contains space types for k-NN similarity search
  */
 public enum SpaceTypes {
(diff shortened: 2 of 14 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java
```diff
@@ -26,6 +26,6 @@
     public static final String LINF = "linf";
     public static final String COSINESIMIL = "cosinesimil";
     public static final String HAMMING_BIT = "hammingbit";
-    public static final String NEGDOTPROD = "negdotprod";
+    public static final String INNER_PROD = "inner_product";
```
jmazanec15: Previously, we didn't put "\_" (i.e. "hammingbit"). However, I like the "\_" and I think it should be kept. Just wanted to add comment here if anyone has strong opinions against this.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/SpaceTypes.java
```diff
@@ -15,41 +15,52 @@
 
 package com.amazon.opendistroforelasticsearch.knn.index;
 
+import java.util.HashMap;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
 
 /**
  * Enum contains space types for k-NN similarity search
```
jmazanec15: Could you update comment to describe why there are 2 values for each space type?

jmazanec15 (commented):
Thanks for submitting this @chenqi0805 . A couple minor comments. Also, when you merge, could you squash first?

src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java
vamshin: Good catch. If this is not exposed to customer, then I like the idea of adding "_"

chenqi0805: Sorry, I am a little confused. Which one do we want to expose to customer? 

```
"space_type": "inner_product"
```
or
```
"space_type": "innerproduct"
```

vamshin: If these values are exposed to customer, my suggestion is to keep it consistent with original ones with out underscore i.e innerproduct.  But if these values are just used internally, then we can use underscore and probably rename others as well(may be part of other Cr)

vamshin (approved):
LGTM! Thanks for the contribution.

jmazanec15 (approved):
LGTM! Thanks for contributing this

src/main/java/com/amazon/opendistroforelasticsearch/knn/plugin/script/KNNScoringUtil.java
```diff
@@ -316,9 +316,9 @@ public static float negdotprod(float[] queryVector, float[] inputVector) {
      *
      * @param queryVector query vector
      * @param docValues   script doc values
-     * @return negdotprod score
+     * @return inner product score
      */
-    public static float negdotprod(List<Number> queryVector, KNNVectorScriptDocValues docValues) {
-        return negdotprod(toFloat(queryVector), docValues.getValue());
+    public static float innerProd(List<Number> queryVector, KNNVectorScriptDocValues docValues) {
```
VijayanB: Shall we name this as innerProduct? Similar to cosinesimil = cosineSimilarity()

VijayanB (approved):
Thanks for making those changes

Issue #211, opened 2020-09-03
Validation check to ensure query vector dimension is not greater than the data vectors


vamshin:
Elasticsearch process crashes when query vector dimension is greater than data vector dimension. Crash possibly happening from nmslib library.

MiguelPF:
Same here, we get the following from the jvm:

```
#
# A fatal error has been detected by the Java Runtime Environment:
#
#  SIGSEGV (0xb) at pc=0x00007fc5f18b11c8, pid=1, tid=239
#
# JRE version: OpenJDK Runtime Environment AdoptOpenJDK (14.0.1+7) (build 14.0.1+7)
# Java VM: OpenJDK 64-Bit Server VM AdoptOpenJDK (14.0.1+7, mixed mode, sharing, tiered, compressed oops, g1 gc, linux-amd64)
# Problematic frame:
# V  [libjvm.so+0xe4d1c8][thread 301 also had an error]
  SymbolTable::do_lookup(char const*, int, unsigned long)+0xb8
#
# Core dump will be written. Default location: Core dumps may be processed with "/usr/lib/systemd/systemd-coredump %P %u %g %s %t %c %h" (or dumping to /usr/share/elasticsearch/core.1)
#
# An error report file with more information is saved as:
# �
[thread 265 also had an error]
*** Error in `/opt/jdk/bin/java': munmap_chunk(): invalid pointer: 0x00007fc5ec0015c0 ***
*** Error in `/opt/jdk/bin/java': free(): invalid pointer: 0x00000000ffffffff ***
*** Error in `/opt/jdk/bin/java': free(): invalid pointer: 0x00000000ffffffff ***
[thread 275 also had an error]
[thread 261 also had an error]
[thread 262 also had an error]
[thread 233 also had an error]
[thread 270 also had an error]
[thread 271 also had an error]
[thread 238 also had an error]
[thread 269 also had an error]
[thread 268 also had an error]
[thread 294 also had an error]
[thread 252 also had an error]


```

PR #332, opened 2021-03-01
Add dimension validation to ANN QueryBuilder

jmazanec15:
*Issue #, if available:*
#211 

*Description of changes:*
Adds a dimension validation check in QueryBuilder class. This change catches invalid query dimensions sooner to prevent potential crash in JNI.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNQueryBuilder.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNQueryBuilder.java
@@ -192,14 +192,21 @@ public class KNNQueryBuilder extends AbstractQueryBuilder<KNNQueryBuilder> {
 
     @Override
     protected Query doToQuery(QueryShardContext context) throws IOException {
+        if (((KNNVectorFieldMapper.KNNVectorFieldType) context.fieldMapper(this.fieldName)).getDimension()
+                != vector.length) {
+            throw new IllegalArgumentException("Invalid vector query vector: " + vector.length +
+                    ". Dimension should be: " +
+                    ((KNNVectorFieldMapper.KNNVectorFieldType) context.fieldMapper(this.fieldName)).dimension);
+        }
+
         return new KNNQuery(this.fieldName, vector, k, context.index().getName());
     }
 
     @Override
     protected boolean doEquals(KNNQueryBuilder other) {
         return Objects.equals(fieldName, other.fieldName) &&
-                       Objects.equals(vector, other.vector) &&
-                       Objects.equals(k, other.k);
+                Objects.equals(vector, other.vector) &&
+                Objects.equals(k, other.k);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNQueryBuilder.java
```diff
@@ -192,6 +194,20 @@ public void doXContent(XContentBuilder builder, Params params) throws IOExceptio
 
     @Override
     protected Query doToQuery(QueryShardContext context) throws IOException {
+
+        MappedFieldType mappedFieldType = context.fieldMapper(this.fieldName);
+
+        if (!(mappedFieldType instanceof KNNVectorFieldMapper.KNNVectorFieldType)) {
+            throw new IllegalArgumentException("Field '" + this.fieldName + "' is not knn type.");
```
VijayanB: Can we say not Knn_vector type?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNQueryBuilder.java
```diff
@@ -192,6 +194,20 @@ public void doXContent(XContentBuilder builder, Params params) throws IOExceptio
 
     @Override
     protected Query doToQuery(QueryShardContext context) throws IOException {
+
+        MappedFieldType mappedFieldType = context.fieldMapper(this.fieldName);
```
VijayanB, line 198: Q: this will never be null right?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNQueryBuilder.java
jmazanec15: Good catch

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNQueryBuilder.java
jmazanec15: Correct

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNQueryBuilder.java
```diff
@@ -192,6 +194,20 @@ public void doXContent(XContentBuilder builder, Params params) throws IOExceptio
...
+
+        if (!(mappedFieldType instanceof KNNVectorFieldMapper.KNNVectorFieldType)) {
```
vamshin, line 200: when will this case arise?

jmazanec15: I dont think we check before this the MappedFieldType of the field. I thought this should be called before we call 204.

VijayanB (approved):

vamshin (approved):
LGTM! Thanks for fixing this.

PR #337, opened 2021-03-12
Change score normalization for negative raw scores

jmazanec15:
*Issue #, if available:*

*Description of changes:*
This PR changes the way in which we convert negative scores to positive scores for the inner product space.

Currently, we calculate it using the following function, where score maps to [-Float.MAX_VALUE, Float.MAX_VALUE]
```
        if (score >= 0)
            return 1 / (1 + score);
        return 2 + ( 1 / (score - 1) );
}
```
This function maps the score to [0, 2], where -Float.MAX_VALUE -> 2 and Float.MAX_VALUE -> 0. The problem with this function is that when score is a large negative number (say 20 million), ` ( 1 / (score - 1)` becomes very small. Then adding the number to 2 results in 2 instead of 1.999, leading to inaccuracy.

In this revision, we change the score function:
```
        if (score >= 0)
            return 1 / (1 + score);
        return -score + 1;
```

Now, if score is very large, it will still hold its value.

This approach still has the weakness that if score is very small (i.e. 0.000000000001), the results may just become 1. To prevent this, we could perform a more precise mapping of [-Float.MAX_VALUE, Float.MAX_VALUE] to [0, Float.MAX_VALUE] by shifting exponent of float by 1 and using most significant bit of exponent to determine "sign" and then setting sign bit to positive. However, this approach probably over complicates things and not be necessary.

### DIFF
--- a/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java
+++ b/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java
@@ -127,7 +127,7 @@ public class KNNWeight extends Weight {
     public static float normalizeScore(float score) {
         if (score >= 0)
             return 1 / (1 + score);
-        return 2 + ( 1 / (score - 1) );
+        return -score + 1;
     }
 }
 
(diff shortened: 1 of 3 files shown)


vamshin (approved):
LGTM! Thanks for the fix.

chenqi0805 (approved):
Thanks for the review invitation. LGTM!

VijayanB (approved):

PR #339, opened 2021-03-19
Expose getValue method from KNNScriptDocValues

VijayanB:
Since user might be interested in accessing the values directly,
allow getValue method. 
To use these in a script, leave out the get prefix and use value as a member variable.

### DIFF
--- a/src/main/resources/com/amazon/opendistroforelasticsearch/knn/plugin/script/knn_whitelist.txt
+++ b/src/main/resources/com/amazon/opendistroforelasticsearch/knn/plugin/script/knn_whitelist.txt
@@ -14,6 +14,7 @@
 # Painless definition of classes used by knn plugin
 
 class com.amazon.opendistroforelasticsearch.knn.index.KNNVectorScriptDocValues {
+  float[] getValue()
 }
 static_import {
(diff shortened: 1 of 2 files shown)


src/main/resources/com/amazon/opendistroforelasticsearch/knn/plugin/script/knn_whitelist.txt
```diff
@@ -14,6 +14,7 @@
 # Painless definition of classes used by knn plugin
 
 class com.amazon.opendistroforelasticsearch.knn.index.KNNVectorScriptDocValues {
```
jmazanec15, line 16: What are error cases?

jmazanec15 (commented):
We will also need to update documentation with this change.

src/main/resources/com/amazon/opendistroforelasticsearch/knn/plugin/script/knn_whitelist.txt
VijayanB: Added additional test cases

VijayanB:
> We will also need to update documentation with this change.

We will be releasing this feature in next release. I will create a task to make sure it is not slipped out.

vamshin (approved):
LGTM! Thanks for adding ability to read the vector from doc values. This gives more control to play with the vectors in painless scripting/custom scoring.

jmazanec15 (approved):

PR #344, opened 2021-04-20
faiss interface refactoring to support multiple methods

jmazanec15:
*Issue #, if available:*
#225 

*Description of changes:*
This PR focuses on refactoring current faiss-support branch's interface to support several additional features including:
1. IVF index type - a cell probe based method that allows a user to reduce search space using a k-Means clustering algorithm. It takes "ncentroids" and "nprobes" as parameters
2. Product quantization - a method to encode vectors to reduce size. It takes "code_size" as a parameter
3. Composite indices - the ability to combine different faiss features into a single index

The interface looks like:
```
{
   "my_vector":{
      "type":"knn_vector",
      "dimension":4,
      "method":{
         "name":"ivf",
         "engine":"faiss",
         "coarse_quantizer":{
            "name":"ivf",
            "parameters":{
               "ncentroids":15
            }
         },
         "encoder":{
            "name":"pq",
            "parameters":{
               "code_size":8
            }
         },
         "parameters":{
            "ncentroids":128
         }
      }
   }
}
```

The main logic where the interface has been refactored can be found in:
1. [KNNVectorFieldMapper](https://github.com/jmazanec15/k-NN/blob/faiss-v1/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java) - where the parsing between the user provided method and the the plugin occurs
2. [KNNMethodContext](https://github.com/jmazanec15/k-NN/blob/faiss-v1/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNMethodContext.java) - stored structure of the user provided method configuration
3. [KNNMethod](https://github.com/jmazanec15/k-NN/blob/faiss-v1/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNMethodContext.java) - structure of a given method supported by a particular engine
4. [KNNLibrary](https://github.com/jmazanec15/k-NN/blob/faiss-v1/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNMethodContext.java) - interface for a particular library. Includes implementations for nmslib and faiss
5. [KNNEngine](https://github.com/jmazanec15/k-NN/blob/faiss-v1/src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNMethodContext.java) - enum mapping name to KNNLibrary

A lot of code was changed in order to support these additional features:
1. Because we use faiss's index factory, only a certain portion of the parameters are configured through the index factory string description. To support additional parameters (for example, ef_construction for HNSW), this PR adds functionality to pass an extra parameter map to the jni to be parsed.
2. Because IVF and PQ require training, in the JNI save index function, this PR implements a training approach where a subset of the data to be indexed is used for training. This is inherently inefficient because it requires each segment to be trained before it can add data to it. In the future, we will introduce a train api that trains before indexing, to work around this. 
3. Several other minor changes to make refactor cleaner/easier

*Testing*
For testing, this PR focuses on addings tests that exercise the interface as opposed to adding end to end tests testing each jni libraries functionality. This is because that functionality will change in the future. Right now, it is just a place holder to get the interface functionality working. That being said, the following test refactoring was done:
1. Added additional unit tests to test faiss interface
2. Refactored old tests so that gradle build passes

*Future Development*
1. Introduce training api
2. Add additional end to end tests
3. Investigate storing data exclusively with faiss (as opposed to storing vectors in doc values in Lucene)

*Notes*
We are in the process of migrating from ODFE to [OpenSearch](https://github.com/opensearch-project/k-NN). Included in this will be porting over the faiss-support branch to OpenSearch. Because porting requires significant refactoring, we will merge this PR and then port the faiss-support branch to OpenSearch.


src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
```diff
@@ -1,34 +1,66 @@
...
 
-public abstract class KNNIndex implements AutoCloseable{
+    protected volatile boolean isClosed = false;
+    protected final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
 
+    protected final long indexPointer;
+    protected final long indexSize;
+    protected final SpaceType spaceType;
 
-    public KNNIndex() {}
```
VijayanB, line 9: you don't need?
```
protected KNNIndex() {}
```

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
```diff
@@ -1,34 +1,66 @@
...
-     * @param ids ids
-     * @param data data
-     * @param indexPath path to save
-     * @param algoParams algorithms params passed to KNN
-     * @param spaceType space type for this index
-     * @param engine verify engine
-     */
-    public abstract void saveIndex(int[] ids, float[][] data, String indexPath, String[] algoParams,
-                                   String spaceType, KNNEngine engine);
+    public KNNQueryResult[] queryIndex(final float[] query, final int k) throws RuntimeException {
```
VijayanB: nit: may be just query? since class itself is called KNNIndex.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
```diff
@@ -38,7 +70,51 @@ public abstract void saveIndex(int[] ids, float[][] data, String indexPath, Stri
...
+    public SpaceType getSpaceType() {
+        return this.spaceType;
+    }
+
+    public void close() {
+        Lock writeLock = readWriteLock.writeLock();
+        writeLock.lock();
+        // Autocloseable documentation recommends making close idempotent. We don't expect to doubly close
+        // but this will help prevent a crash in that situation.
+        if (this.isClosed) {
```
VijayanB: you don't need to unlock?

jmazanec15: Good question. The unlock happens in the finally block, so no.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
jmazanec15: You're right. Index is redundant. I will refactor.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
jmazanec15: No I dont think so. The constructor for the index is only needed in the KNNFaissIndex and KNNNmslibIndex constructors, which are only used in their respective loadIndex static functions. 

What do you think?

src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java
```diff
@@ -21,12 +21,33 @@
...
-    public static final String L1 = "l1";
-    public static final String LINF = "linf";
-    public static final String COSINESIMIL = "cosinesimil";
-    public static final String HAMMING_BIT = "hammingbit";
-    public static final String INNER_PROD = "innerproduct";
     public static final String DIMENSION = "dimension";
-    public static final String KNNEngine = "knnEngine";
+    public static final String KNN_ENGINE = "engine";
+    public static final String KNN_METHOD= "method";
+    public static final String EXTRA_PARAMETERS = "extraParameters";
```
vamshin: Nit: why not call ```extra_parameters``` to keep it consistent?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
```diff
@@ -38,7 +70,51 @@ public abstract void saveIndex(int[] ids, float[][] data, String indexPath, Stri
...
+        if (indexPath == null || indexPath.isEmpty()) {
+            return 0;
+        }
+        File file = new File(indexPath);
+        if (!file.exists() || !file.isFile()) {
+            return 0;
+        }
 
-    public abstract void close();
+        return file.length() / 1024 + 1;
```
vamshin: minor: ```return  (file.length() / 1024) + 1``` for readability?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -137,43 +168,58 @@ public Builder(String name, String spaceType, String m, String efConstruction, S
...
         @Override
         public KNNVectorFieldMapper build(BuilderContext context) {
-            if (this.spaceType == null) {
-                this.spaceType = getSpaceType(context.indexSettings());
-            }
-
-            if (this.m == null) {
-                this.m = getM(context.indexSettings());
-            }
+            // Originally, a user using nmslib would set spaceType, m and efConstruction with index settings. We are in
```
vamshin, line 172: This function need to be tested manually for clusters moving from 7.10.2 with out faiss to 7.10.2 with faiss

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -268,25 +315,39 @@ public int getDimension() {
...
+        this.fieldType.putAttribute(KNN_ENGINE, knnEngine);
+        this.fieldType.putAttribute(SPACE_TYPE, spaceType);
+        this.fieldType.putAttribute(MINIMUM_DATAPOINTS, knnMethod.getTrainingDatasetSizeLimit().toString());
+        this.fieldType.putAttribute(MINIMUM_DATAPOINTS, knnMethod.getMinimumDatapoints().toString());
+        this.fieldType.putAttribute(KNN_METHOD, knnMethod.generateMethod());
+
+        if (KNNEngine.FAISS.getName().equals(knnEngine)) {
+            this.fieldType.putAttribute(KNNConstants.EXTRA_PARAMETERS, knnMethod.generateExtraParameters());
+        }
+
```
vamshin: We need to test for race condition with having multiple indices with different engine type and algorithms

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -81,55 +91,78 @@ public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer)
...
             Map<String, String> fieldAttributes = field.attributes();
-            String engineName = fieldAttributes.getOrDefault(KNNConstants.KNNEngine, KNNSettings.INDEX_KNN_DEFAULT_ENGINE);
-            String spaceType = SpaceTypes.getValueByKey(fieldAttributes.getOrDefault(KNNConstants.SPACE_TYPE, SpaceTypes.l2.getKey()));
+            String engineName = fieldAttributes.getOrDefault(KNNConstants.KNN_ENGINE, KNNEngine.DEFAULT.getName());
+            String spaceType = fieldAttributes.getOrDefault(KNNConstants.SPACE_TYPE, SpaceType.L2.getValue());
+            int trainingDatasetSizeLimit = Integer.parseInt(fieldAttributes.getOrDefault(
+                    KNNConstants.TRAINING_DATASET_SIZE_LIMIT,
+                    KNNMethodContext.DEFAULT_TRAINING_DATASET_SIZE_LIMIT.toString()));
+            int minimumDatapoints = Integer.parseInt(fieldAttributes.getOrDefault(
+                    KNNConstants.MINIMUM_DATAPOINTS, KNNMethodContext.DEFAULT_MINIMUM_DATAPOINTS.toString()));
```
vamshin, line 100: These all would go away with segment level training right?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java
```diff
@@ -88,39 +91,52 @@ public String getIndexName() {
...
-        }
-        return hnswFiles;
-    }
+            String fileExtension = reader.getSegmentInfo().info.getUseCompoundFile()
+                    ? knnEngine.getCompoundExtension() : knnEngine.getExtension();
 
-    private ShardPath shardPath() {
-        return indexShard.shardPath();
-    }
+            for (FieldInfo fieldInfo : reader.getFieldInfos()) {
```
vamshin, line 126: warmup api tests should be covered with multiple knn fields and different Engine types in the same index

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
```diff
@@ -268,25 +315,39 @@ public int getDimension() {
...
-        this.fieldType.putAttribute(KNNConstants.KNNEngine, knnEngine);
+
+        if (KNNEngine.NMSLIB.getName().equals(knnEngine)) {
+            this.fieldType.putAttribute(KNNConstants.HNSW_ALGO_M, m);
+            this.fieldType.putAttribute(KNNConstants.HNSW_ALGO_EF_CONSTRUCTION, efConstruction);
+        }
+
+        this.fieldType.putAttribute(KNN_ENGINE, knnEngine);
+        this.fieldType.putAttribute(SPACE_TYPE, spaceType);
+        this.fieldType.putAttribute(MINIMUM_DATAPOINTS, knnMethod.getTrainingDatasetSizeLimit().toString());
```
vamshin: should be ```TRAINING_DATAPOINTS```

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
VijayanB: @jmazanec15 but that will only happen if you are returning inside from try. Hence, if isClosed == true, your unlock will not be called.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
VijayanB: Got it. if it is not used anywhere, then we don't have to add it.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
```diff
@@ -1,34 +1,66 @@
...
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
-import com.amazon.opendistroforelasticsearch.knn.index.util.KNNEngine;
+/**
+ * Encapsulates functionality to interface with an underlying k-NN Engine
+ */
+public abstract class KNNIndex implements AutoCloseable {
 
-public abstract class KNNIndex implements AutoCloseable{
+    protected volatile boolean isClosed = false;
```
VijayanB: all protected member variables can be made private

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
```diff
@@ -38,7 +70,51 @@ public abstract void saveIndex(int[] ids, float[][] data, String indexPath, Stri
...
+            writeLock.unlock();
+        }
+    }
+
+    /**
+     * determines the size of the engine index on disk
+     * @param indexPath absolute path of the index
+     *
+     */
+    protected static long computeFileSize(String indexPath) {
```
VijayanB: This looks like a utility method. we can keep this class have only required methods, what do you think?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
VijayanB: let's create constant to understand what does 1024 is and why add 1? Also, method can be named as getSizeInKB(String filePath)

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
```diff
@@ -38,7 +70,51 @@ public abstract void saveIndex(int[] ids, float[][] data, String indexPath, Stri
...
-    public abstract void close();
+        return file.length() / 1024 + 1;
+    }
+
+
+    /*
+     * Wrappers around Jni functions
+     */
+    protected abstract KNNQueryResult[] queryIndexJniWrapper(long indexPointer, float[] query, int k);
+    protected abstract void gcJniWrapper(long indexPointer);
```
VijayanB: gcJNIWrapper

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNMethod.java
```diff
@@ -0,0 +1,271 @@
...
+     */
+    public KNNMethod(String name, Set<SpaceType> spaces, Map<String, Parameter<?>> parameters,
+                     Map<String, MethodComponent> encoders, boolean isCoarseQuantizerAvailable) {
+        this.methodComponent = new MethodComponent(name, parameters);
+        this.spaces = spaces;
+        this.encoders = encoders;
+        this.isCoarseQuantizerAvailable = isCoarseQuantizerAvailable;
+    }
+
+    private MethodComponent methodComponent;
```
VijayanB: shall we move declaration above?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNMethod.java
```diff
@@ -0,0 +1,271 @@
...
+     * @return map of extra parameters
+     */
+    public Map<String, Object> generateExtraParameterMap(KNNMethodContext knnMethodContext) {
+        return methodComponent.generateExtraParameterMap(knnMethodContext.getMethodComponent().getParameters());
+    }
+
+    /**
+     * MethodComponent defines the structure of an individual component that can make up an index
+     */
+    public static class MethodComponent {
```
VijayanB: Have you consider using Builder pattern?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNMethod.java
```diff
@@ -0,0 +1,271 @@
...
+            }
+        }
+    }
+
+    /**
+     * Parameter that can be set for a method component
+     *
+     * @param <T> Type parameter takes
+     */
+    public static abstract class Parameter<T> {
```
VijayanB: if it is public, may be can extract it into new file. What do you think?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
jmazanec15: I believe it will still be reached: https://docs.oracle.com/javase/tutorial/essential/exceptions/finally.html.

Take this sample program:
```
import java.util.*;
import java.io.IOException;
class test {

    public static void main(String[] args)
    {
	try {
		System.out.println("Trying");
		throw new IOException("Test exception");
	} catch (Exception ex) {
		throw new RuntimeException("Test exception: " + ex);
	} finally {
		System.out.println("Reached finally");
	}
    }
}
```

The output is:
```
$ java test
Trying
Reached finally
Exception in thread "main" java.lang.RuntimeException: Test exception: java.io.IOException: Test exception
	at test.main(test.java:11)
```

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
jmazanec15: Good point. Will fix.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
jmazanec15: You are right. This can be moved out of this class to reduce dependencies. I will move it out.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
jmazanec15: Good point. Will update.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
jmazanec15: Will change.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNMethod.java
jmazanec15: Yes, that makes sense.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNMethod.java
jmazanec15: Let me think about this. I think it could be used.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNMethod.java
jmazanec15: This may not be required to be public. Let me think about it.

src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java
jmazanec15: Will fix.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
jmazanec15: Will fix

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
jmazanec15: Will note for future ref

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
jmazanec15: Correct.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexShard.java
jmazanec15: Yes, will address these in a future PR.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
jmazanec15: Good catch. Will fix. This will be removed in future.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNVectorFieldMapper.java
jmazanec15: Yes, will add more extensive ITs once training approach gets finalized.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNMethod.java
jmazanec15: Added it in the latest revision for KNNMethod and MethodComponent

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNMethod.java
jmazanec15: Kept it public and moved it out of the file

src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java
```diff
@@ -21,12 +21,33 @@
...
     public static final String HNSW_ALGO_INDEX_THREAD_QTY = "indexThreadQty";
-    public static final String L2 = "l2";
-    public static final String L1 = "l1";
-    public static final String LINF = "linf";
-    public static final String COSINESIMIL = "cosinesimil";
-    public static final String HAMMING_BIT = "hammingbit";
-    public static final String INNER_PROD = "innerproduct";
     public static final String DIMENSION = "dimension";
-    public static final String KNNEngine = "knnEngine";
+    public static final String KNN_ENGINE = "engine";
```
VijayanB, line 25: for future: lets sort constants based on alphabets. This will help readability and easy to search specific constant.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java
```diff
@@ -422,22 +400,12 @@ public void setClusterService(ClusterService clusterService) {
     static class SpaceTypeValidator implements Setting.Validator<String> {
 
         @Override public void validate(String value) {
-            if (value == null || !SpaceTypes.contains(value.toLowerCase())){
+            if (value == null || !SpaceType.getValues().contains(value.toLowerCase())){
                 throw new InvalidParameterException(String.format("Unsupported space type: %s", value));
```
VijayanB: if this is user facing message, can we add what are supported types as well? This will give more information to the user. What do you think?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java
```diff
@@ -74,57 +74,66 @@ public Scorer scorer(LeafReaderContext context) throws IOException {
...
 
-            if(hnswFiles.isEmpty()) {
-                logger.debug("[KNN] No hnsw index found for field {} for segment {}",
+            if(engineFiles.isEmpty()) {
+                logger.debug("[KNN] No engine index found for field {} for segment {}",
                         knnQuery.getField(), reader.getSegmentName());
                 return null;
             }
 
             FieldInfo queryFieldInfo = reader.getFieldInfos().fieldInfo(knnQuery.getField());
```
VijayanB: This is similar to line 77. can we reuse it?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java
```diff
@@ -74,57 +74,66 @@ public Scorer scorer(LeafReaderContext context) throws IOException {
...
+            Path indexPath = PathUtils.get(directory, engineFiles.get(0));
             final KNNQueryResult[] results;
+            final KNNIndex index = knnIndexCache.getIndex(indexPath.toString(), knnQuery.getIndexName(), spaceType);
 
-            final KNNIndex index = knnIndexCache.getIndex(indexPath.toString(), knnQuery.getIndexName());
-            //we check instance of KNNINdex
-            if (fieldAttributes.containsValue(KNNEngine.NMSLIB.getKnnEngineName())) {
-                assert(index instanceof KNNNmsLibIndex);
-                results = index.queryIndex(
+            if ((fieldAttributes.containsValue(KNNEngine.NMSLIB.getName()) && index instanceof KNNNmsLibIndex)
```
VijayanB: flip if condition to throw exception if it is not matched.
In general: if there is an exception thrown in else, you can flip if to have that part of inside if statement and everything else outside if statement. This will reduce nested level.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java
```diff
@@ -74,57 +74,66 @@ public Scorer scorer(LeafReaderContext context) throws IOException {
...
+            /*
              * Scores represent the distance of the documents with respect to given query vector.
              * Lesser the score, the closer the document is to the query vector.
              * Since by default results are retrieved in the descending order of scores, to get the nearest
              * neighbors we are inverting the scores.
              */
+            if (results.length == 0) {
+                logger.debug("[KNN] Query yielded 0 results");
+                return null;
+            } else {
```
VijayanB: else is redundant

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java
```diff
@@ -0,0 +1,136 @@
...
+
+    private String name;
+    private Map<String, Parameter<?>> parameters;
+
+    /**
+     * Constructor
+     *
+     * @param builder to build method component
+     */
+    private MethodComponent(Builder builder) {
```
VijayanB, line 38: null check for builder

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    /**
+     * Validate that the methodComponentContext is a valid configuration for this methodComponent
+     *
+     * @param methodComponentContext to be validated
+     */
+    public void validate(KNNMethodContext.MethodComponentContext methodComponentContext) {
+        Map<String, Object> providedParameters = methodComponentContext.getParameters();
+        if (providedParameters != null) {
```
VijayanB: if providedParameters == null
   return

VijayanB: Nested is not good for readability and error prone.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        /**
+         * Add parameter entry to parameters map
+         *
+         * @param parameterName name of the parameter
+         * @param parameter parameter to be added
+         * @return this builder
+         */
+        public Builder putParameter(String parameterName, Parameter<?> parameter) {
```
VijayanB: may be addParameter?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java
```diff
@@ -0,0 +1,136 @@
...
+        private String name;
+        private Map<String, Parameter<?>> parameters;
+
+        /**
+         * Method to get a Builder instance
+         *
+         * @param name of method component
+         * @return Builder instance
+         */
+        public static MethodComponent.Builder builder(String name) {
```
VijayanB: you can skip MethodComponent and refer as Builder

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/Parameter.java
```diff
@@ -0,0 +1,86 @@
...
+    private boolean inMethodString;
+    protected Function<T, Boolean> validator;
+
+    /**
+     * Constructor
+     *
+     * @param defaultValue of the parameter
+     * @param inMethodString whether the parameter is included in method string
+     */
+    public Parameter(T defaultValue, boolean inMethodString, Function<T, Boolean> validator) {
```
VijayanB: Why not ```Predicate <T>``` instead of ```Function<T, Boolean>```

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java
```diff
@@ -81,55 +91,78 @@ public void addKNNBinaryField(FieldInfo field, DocValuesProducer valuesProducer)
...
             // Pass the path for the nms library to save the file
             String tempIndexPath = indexPath + TEMP_SUFFIX;
             AccessController.doPrivileged(
-                    new PrivilegedAction<Void>() {
-                        public Void run() {
-                            if(faissindex) {
-                                KNNFaissIndex.saveIndex(pair.docs, pair.vectors, tempIndexPath, algoParams, spaceType);
-                            } else {
-                                KNNNmsLibIndex.saveIndex(pair.docs, pair.vectors, tempIndexPath, algoParams, spaceType);
+                    (PrivilegedAction<Void>) () -> {
```
VijayanB, line 138: please avoid else/else if , it is difficult to follow.

VijayanB:
```
            AccessController.doPrivileged(
                    (PrivilegedAction<Void>) () -> {
                        if(KNNEngine.NMSLIB.getName().equals(knnEngine.getName())) {
                            KNNNmsLibIndex.save(pair.docs, pair.vectors, tempIndexPath, algoParams, spaceType,
                                    method);
                            return null;
                        }
                        if (KNNEngine.FAISS.getName().equals(knnEngine.getName())) {
                            String extraParametersString = fieldAttributes.getOrDefault(KNNConstants.EXTRA_PARAMETERS, null);
                            Map<String, Object> extraParameterMap = Collections.emptyMap();
                            if (extraParametersString != null) {
                                try {
                                    extraParameterMap = XContentFactory.xContent(XContentType.JSON).createParser(NamedXContentRegistry.EMPTY, DeprecationHandler.THROW_UNSUPPORTED_OPERATION, extraParametersString).map();
                                } catch (IOException e) {
                                    throw new IllegalStateException(e);
                                }

                            }

                            KNNFaissIndex.save(pair.docs, pair.vectors, tempIndexPath, extraParameterMap, spaceType,
                                    method, trainingDatasetSizeLimit, minimumDatapoints);
                            return null;
                        } 
                        throw new IllegalStateException("Invalid engine: "+knnEngine.getName());
                    }
            );

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java (hunk 17 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java:25]: Agree. Do you think that it makes sense to localize constants to specific files or keeping all constants in a single class?

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java (hunk 18 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNSettings.java]: Yes, I can add that.

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java (hunk 19 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java]: You're right, I will fix this

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java (hunk 20 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java]: that makes sense. Will fix.

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java (hunk 21 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNWeight.java]: Right, I dont need that debug statement anyway.

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java (hunk 22 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java:38]: My thinking is that because the component is built from the builder and is private, a null check is not needed. What do you think?

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java (hunk 23 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java]: Good point.

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java (hunk 24 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java]: My thinking was that since it is adding a parameter into a map, it should be similar to Map.put. What do you think?

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java (hunk 25 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java]: Good catch

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/Parameter.java (hunk 26 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/Parameter.java]: I did not know about Predicate. I will replace that.

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java (hunk 27 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/codec/KNN80Codec/KNN80DocValuesConsumer.java:138]: Sure, will update.

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java (hunk 22 above)
Comment (VijayanB) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java:38]: Yes. You are right.

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java (hunk 24 above)
Comment (VijayanB) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java]: but parameter map implementation is abstracted by Builder. So from Builder point of you we are adding parameter, but internally we are using map to insert.

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java (hunk 17 above)
Comment (VijayanB) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java:25]: I prefer single file, it is easy to maintain.

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java (hunk 24 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/index/MethodComponent.java]: You are right. I will fix

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java (hunk 17 above)
Comment (jmazanec15) [reply src/main/java/com/amazon/opendistroforelasticsearch/knn/common/KNNConstants.java:25]: Makes sense. We have some what of a hybrid approach right now. But we can adjust in the future

File: src/main/java/com/amazon/opendistroforelasticsearch/knn/index/IndexUtil.java [hunk 28]
```diff
@@ -0,0 +1,41 @@
...
+ *   permissions and limitations under the License.
+ */
+
+package com.amazon.opendistroforelasticsearch.knn.index;
+
+import java.io.File;
+
+public class IndexUtil {
+
+    public static Long BYTES_PER_KILOBYTES = 1024L;
```
VijayanB: this can be private

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/IndexUtil.java
```diff
@@ -0,0 +1,41 @@
...
+
+    /**
+     * Determines the size of a file on disk in kilobytes
+     *
+     * @param filePath path to the file
+     * @return length of the file in kilobytes
```
VijayanB: file size in KB

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/IndexUtil.java
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public static long getFileSizeInKB(String filePath) {
+        if (filePath == null || filePath.isEmpty()) {
```
VijayanB, line 31: This will not differentiate empty file with invalid file path or null. Is this intended?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
```diff
@@ -1,34 +1,64 @@
...
 
-import com.amazon.opendistroforelasticsearch.knn.index.util.KNNEngine;
+/**
+ * Encapsulates functionality to interface with an underlying k-NN Engine
+ */
+public abstract class KNNIndex implements AutoCloseable {
 
-public abstract class KNNIndex implements AutoCloseable{
+    private volatile boolean isClosed = false;
+    private final ReadWriteLock readWriteLock = new ReentrantReadWriteLock();
```
VijayanB: can this be moved inside constructor?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
```diff
@@ -1,34 +1,64 @@
...
-     */
-    public abstract void saveIndex(int[] ids, float[][] data, String indexPath, String[] algoParams,
-                                   String spaceType, KNNEngine engine);
-
+    public KNNQueryResult[] query(final float[] query, final int k) throws RuntimeException {
+        Lock readLock = readWriteLock.readLock();
+        readLock.lock();
+        try {
+            if (this.isClosed) {
+                throw new IOException("Index is already closed");
```
VijayanB: Something like " Cannot query closed Index". 
"Index is already closed" makes sense inside close method. What do you think?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -314,14 +314,17 @@ public KNNIndexCacheEntry loadIndex(String indexPathUrl, String indexName) throw
         final WatcherHandle<FileWatcher> watcherHandle = resourceWatcherService.add(fileWatcher);
 
         // loadIndex from different library
-        if (KNNSettings.getKnnEngine(indexName).contains(KNNEngine.FAISS.getKnnEngineName())) {
-            final KNNFaissIndex knnIndex = KNNFaissIndex.loadIndex(indexPathUrl, getQueryParams(indexName), KNNSettings.getSpaceType(indexName));
-            return new KNNIndexCacheEntry(knnIndex, indexPathUrl, indexName, watcherHandle);
+        final KNNIndex knnIndex;
 
+        if (indexPathUrl.contains(KNNEngine.NMSLIB.getExtension())) {
```
VijayanB: map be endsWith?

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
```diff
@@ -314,14 +314,17 @@ public KNNIndexCacheEntry loadIndex(String indexPathUrl, String indexName) throw
...
+            knnIndex = KNNNmsLibIndex.load(indexPathUrl, getQueryParams(indexName), spaceType);
+        } else if (indexPathUrl.contains(KNNEngine.FAISS.getExtension())) {
```
VijayanB: same as 319

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/IndexUtil.java
jmazanec15: Right, will change.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/IndexUtil.java
jmazanec15: Good catch.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/IndexUtil.java
jmazanec15: So I guess it would say an empty file has a size of 1 Kb, where as a non-existent file has a size of 0.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
jmazanec15: good catch, will fix

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndex.java
jmazanec15: yes that makes sense.

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
jmazanec15: So the reason I did contains instead of ends with is that the extension could be a compound extension. However, endswith is more correct, so I will fix

src/main/java/com/amazon/opendistroforelasticsearch/knn/index/KNNIndexCache.java
jmazanec15: Got it

jmazanec15:
Closing PR now. Will continue work on OpenSearch repo.

